PR #40, opened 2015-10-29
Testing fixes + factory for data mocking

akx:


akx:
Rebased this on top of current master.

PR #48, opened 2015-11-04
Fix errors when saving model and duplicate variable

rphlo:


kk/models/base.py
```diff
@@ -28,7 +28,7 @@ def save(self, *args, **kwargs):
             pass
         else:
             raise Exception('Unsupported primary key field: %s' % pk_type)
-        super().save(*args, **kwargs)
+        super(ModifiableModel, self).save(*args, **kwargs)
```
akx: This shouldn't be necessary if you're running Python 3...

akx:
@rphlo: Can you squash the duplicate changes again, please?

```python
git fetch --all
git rebase -i upstream/master
```
(You can rearrange lines in the rebase todo; `squash` and `fixup` always affect the line above.)
:)

piotras:
@rphlo id with default method approach seems to be better than save().

PR #67, opened 2015-11-06
Image fixes (and other incidental stuff)

akx:


kk/models/images.py
```diff
@@ -1,24 +1,16 @@
...
-def get_images_dir():
-    return settings.IMAGES_DIR
-
-file_storage = FileSystemStorage(location=get_images_dir())
-
 
 class Image(ModifiableModel):
     title = models.CharField(verbose_name=_('The title'), max_length=255, blank=True, default='')
     caption = models.TextField(verbose_name=_('Caption'), blank=True, default='')
     height = models.IntegerField(verbose_name=_('Height'), default=0)
```
rphlo: Could add `editable=False`

kk/settings/__init__.py
```diff
@@ -0,0 +1,12 @@
+from .base import *
+
+try:
+    import local_settings
```
rphlo: Add local_settings.py in `.gitignore`?

kk/tests/test_images.py
```diff
@@ -1,193 +1,80 @@
...
+def test_38_get_scenario_with_images(client, default_hearing):
+    """
+    Check images exist in scenario payloads
+    """
+    data = get_data_from_response(client.get(get_hearing_detail_url(default_hearing.id, 'scenarios')))
+    first_scenario = data[0]
+    check_entity_images(first_scenario)
+
+
+@pytest.mark.django_db
```
piotras: Do we need to have db decorator per every method?

PR #72, opened 2015-11-09
kk_mock_populate: Create users and comments

akx:


kk/factories/hearing.py
```diff
@@ -1,11 +1,20 @@
...
 import random
+from django.contrib.auth import get_user_model
 from django.utils.timezone import now
+from kk.factories.comment import BaseCommentFactoryMixin
 from kk.models import Hearing, Label, Scenario
+from kk.models.hearing import HearingComment
+from kk.models.scenario import ScenarioComment
+
+
+def get_random_user():
```
rphlo: defined also in factory utils

kk/factories/hearing.py
```diff
@@ -39,6 +58,10 @@ def post(obj, create, extracted, **kwargs):
             scenario = ScenarioFactory(hearing=obj)
             print(".. Created scenario %s" % scenario.pk)
 
+        for x in range(random.randint(1, 5)):
+            comment = HearingCommentFactory(hearing=obj)
```
rphlo: Missing update of nb_comment field

kk/factories/hearing.py
akx: Copypaste fail. Will fix.

kk/factories/hearing.py
akx: Good catch, adding

PR #83, opened 2015-11-09
Comment refactoring & fixes

akx:


kk/factories/utils.py
```diff
@@ -1,7 +1,12 @@
 # -*- coding: utf-8 -*-
 
 from django.contrib.auth import get_user_model
+from kk.factories.user import UserFactory
 
 
 def get_random_user():
-    return get_user_model().objects.order_by("?").first()
+    users_qs = get_user_model().objects
+    if not users_qs.count():  # Last-ditch attempt to create _some_ random users
```
rphlo: As long as this function is only used in factory for test purpose it's ok...

akx: Yeah. I'm pretty sure no one will attempt to do `get_random_user()` in production code...

kk/models/base.py
```diff
@@ -81,10 +81,10 @@ class WithCommentsMixin(models.Model):
     """
     Mixin for models which can be commented.
     """
-    n_comments = models.IntegerField(verbose_name=_('Number of comments'), blank=True, default=0)
+    n_comments = models.IntegerField(verbose_name=_('Number of comments'), blank=True, default=0, editable=False)
 
     def recache_n_comments(self):
-        new_n_comments = self.comments.count()
+        new_n_comments = self.comments.exclude(deleted=True).count()
```
rphlo: Would be cleaner to use an objects manager that exclude deleted in default queryset

akx: We'll have to do a pass to do that for all API viewsets anyway. I don't think any regular API should return deleted objects.

akx: Created #84 about that.

PR #86, opened 2015-11-10
Exclude soft deleted from default queryset

rphlo:


akx:
Can we get a test case for this? :)

rphlo:
test_n_comments_updates should cover that.

PR #99, opened 2015-11-16
Add unvote and unfollow endpoints

rphlo:


kk/views/comment.py
```diff
@@ -83,3 +83,22 @@ def votes(self, request, **kwargs):
         comment.recache_n_votes()
         # return success
         return response.Response({'status': 'Vote has been added'}, status=status.HTTP_201_CREATED)
+
+    @detail_route(methods=['post'])
+    def unvotes(self, request, **kwargs):
```
akx: This should probably be `unvote` (and likewise, `votes` should be `vote`), actually.

kk/views/hearing.py
```diff
@@ -80,6 +80,17 @@ def follow(self, request, pk=None):
         # return success
         return response.Response({'status': 'You follow a hearing now'}, status=status.HTTP_201_CREATED)
 
+    @detail_route(methods=['post'])
+    def unfollow(self, request, pk=None):
+        hearing = self.get_object()
```
akx: This probably needs an authentication check as well.

rphlo: Since we have `permission_classes = (permissions.IsAuthenticatedOrReadOnly,)` there is already a check for authentication of post request.

kk/views/comment.py
rphlo: True I didn't want to break the pattern... but `vote` and `unvote` are probably better

kk/views/hearing.py
akx: Ah, alright! Didn't notice that.

PR #100, opened 2015-11-16
User API endpoint

rphlo:


kk/urls_v1.py
```diff
@@ -20,4 +21,5 @@
     url(r'^', include(hearing_comments_router.urls, namespace='v1')),
     url(r'^', include(hearing_child_router.urls, namespace='v1')),
     url(r'^', include(scenario_comments_router.urls, namespace='v1')),
+    url(r'^me', UserDataView.as_view()),
```
akx: This should probably be registered in the `router` for all that fancy autodocumentation jazz.

kk/urls_v1.py
```diff
@@ -15,9 +16,13 @@
 scenario_comments_router = routers.NestedSimpleRouter(hearing_child_router, r'scenarios', lookup='comment_parent')
 scenario_comments_router.register(r'comments', ScenarioCommentViewSet, base_name='comments')
 
+user_router = routers.SimpleRouter()
```
akx: Mmm, you could just use the base `router`, couldn't you?

akx:
Sorry for letting this hang for this long!

PR #104, opened 2015-11-23
Refactor data model

akx:
Getting rid of Scenarios in favor of Sections.

(And other improvements and cleanup...)


kk/utils/drf_enum_field.py
```diff
@@ -0,0 +1,36 @@
...
+    def to_internal_value(self, data):
+        # Normal logic:
+        for choice in self.enum_type:
+            if choice.name == data or choice.value == data:
+                return choice
+
+        # Case-insensitive logic:
+        l_data = force_text(data).lower()
+        for choice in self.enum_type:
+            if choice.name.lower() == l_data or force_text(choice.value).lower() == data:
```
rphlo: `force_text(choice.value).lower() == l_data`

akx: Oy, good catch. (As it happens this method is more or less the only one that's not test-covered yet...)

PR #112, opened 2015-11-27
Closed semantics (and other misc. fixes...)

akx:


akx:
(I since changed Coveralls settings to not mark coverage decreases less than 5% as trouble.)

PR #115, opened 2015-12-01
Add WYMEditor for HTML editing.

akx:
I'm not entirely convinced about it though...

Refs #102


akx:
Passing this to @juyrjola because, as the commit message notes, I'm not entirely convinced this is the right tool for the job, though I'd much prefer a WYMish editor instead of a more WYGy one.

Screenie:

image

juyrjola:
Passing it on to @rikuoja who will review the concept.

akx:
Closing this in favor of TinyMCE in #118.

PR #117, opened 2015-12-03
Rename to Democracy; upgrade to Django 1.9; add migrations

akx:


akx:
Oh, right. I just search-and-replaced. Gonna add an autopep8 + isort commit. :)

PR #122, opened 2015-12-07
Enable message internationalization with Transifex

akx:


requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+pytest==2.8.3
```
rphlo: add `-r requirements.txt` so to avoid having to include the default requirements

manage.py
```diff
@@ -1,10 +1,34 @@
...
+        return True
+    return False
+
+
 if __name__ == "__main__":
     os.environ.setdefault("DJANGO_SETTINGS_MODULE", "kerrokantasi.settings")
-
+    if run_npm_script():
+        sys.exit(0)
     from django.core.management import execute_from_command_line
```
rphlo: Is it that if there is some `scripts` in package.json we never run the django command?

akx: No; `run_npm_script()` returns True iif it actually ran a command from package.json.

rphlo: Ok I got it now... :)

requirements-dev.txt
akx: Added.

manage.py
akx: Added a docstring, though. :)

Issue #119, opened 2015-12-07
Add authless preview URLs for non-public hearings


akx:
Allow people in possession of a special keyed URL to have read access to non-public hearings for preview purposes.

PR #123, opened 2015-12-07
Add authless URL for non public hearings, Issue #119

rphlo:


democracy/models/hearing.py
```diff
@@ -43,6 +43,16 @@ def may_comment(self, request):
             raise ValidationError(_("%s is closed and does not allow comments anymore") % self, code="hearing_closed")
         return super().may_comment(request)
 
+    @property
+    def preview_code(self):
+        import base64
```
akx: Using the `hmac` module might be easier:

```python
import hmac
return hmac.new(
    key=settings.SECRET_KEY.encode("utf8"),
    msg=str(self.pk).encode("utf8")
).hexdigest()
```

democracy/views/hearing.py
```diff
@@ -99,8 +100,25 @@ def get_serializer(self, *args, **kwargs):
         kwargs['context'] = self.get_serializer_context()
         return serializer_class(*args, **kwargs)
 
+    @staticmethod
+    def check_preview_code(id, code):
```
akx: The key-generating logic should be refactored into a function that can be shared between here and the model.

democracy/views/hearing.py
```diff
@@ -99,8 +100,25 @@ def get_serializer(self, *args, **kwargs):
...
+        import base64
+        import hashlib
+        check = base64.urlsafe_b64encode(
+            hashlib.sha256(
+                "{}{}".format(id, settings.SECRET_KEY).encode('utf-8')
+            ).digest()
+        ).decode('utf-8').replace('=', '')
+        return check == code
+
     def get_queryset(self):
```
akx: Would there maybe be a neater way to implement this if you overrode `get_object()` instead of `get_queryset()`? After all, `?preview=` should only work for the detail view, not lists etc.

akx:
The only CI failure is

> democracy/utils/hmac_hash.py:9:68: W292 no newline at end of file

and this LGTM otherwise, so merging and fixing that style nit next :)

Thanks!

PR #124, opened 2015-12-08
Admin usability improvements

### LINKED ISSUE #102: As an admin, I want to use some sort of WYSIWYGish editor for content

### LINKED ISSUE #118: Implement usability embetterments for the admin
# General Admin changes

* Use django-chosen or similar for M2M fields
* Hide the public flag from labels
* Default "ordering" fields to 1; add a help text
* Use TinyMCE 4 for rich text editing in the admin

# Image Admin changes

* Hide the public flag from images
* Hide image titles in the admin
* Limit the size of the image caption field

# Hearing Admin changes

* Add https://pypi.python.org/pypi/django-nested-admin and remove Sections from top level
* Limit the length of the `abstract` (and the size of the field accordingly)
* Hide latitude, longitude, borough and servicemap_url
* Hide the public flag for sections
* Move the section identifier field further down
* Prepopulate the first section of hearings as an introduction in the admin
* Hide the Commenting field from section admins; copy the Hearing's commenting value into them on save
* Limit the length of the section `abstract`

akx:
Closes #118 
Closes #102


akx:
Agh, isort apparently skips `__init__.py`s by default -- will re-isort after I fix that.

democracy/admin/widgets.py
```diff
@@ -0,0 +1,80 @@
...
+        return forms.Media(
+            js=[static("admin/hoist-jquery.js"), static("admin/select2/select2.min.js")],
+            css={"all": [static("admin/select2/select2.min.css")]}
+        )
+
+    def render(self, name, value, attrs=None, choices=()):
+        id = attrs.get("id", "id_%s" % name)
+        output = [super(Select2SelectMultiple, self).render(name, value, attrs, choices)]
+        output.append('<script type="text/javascript">addEvent(window, "load", function(e) {')
+        output.append('$("#%s").select2(); });</script>\n' % id)
```
rphlo: Those 2 lines should be merged in one append

rphlo: also `.format` is prefered

democracy/admin/widgets.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def render(self, name, value, attrs=None):
+        config = deepcopy(TINYMCE_CONFIG)
+        output = [super().render(name, value, attrs)]
+        output.append('<script type="text/javascript">addEvent(window, "load", function(e) {')
+        output.append('var config = %s;\n' % json.dumps(config))
+        output.append('config.selector = "#" + document.querySelector("textarea[name=%s]").id;\n' % name)
+        output.append('tinymce.init(config);\n')
+        output.append('});')
+        output.append('</script>')
```
rphlo: similar. One append and `.format`

democracy/views/comment.py
```diff
@@ -60,13 +60,13 @@ def _check_may_comment(self, request):
         parent = self.get_comment_parent()
         try:
             may_comment = parent.may_comment(request)
+            if not may_comment:  # pragma: no cover
```
rphlo: Would be good to check for coverage on this.

akx: Technically yes, but it's a fallback case; right now `may_comment` implementations always return True or raise a ValidationError.

rphlo: The name of the function is confusing as `may` imply True/False answer. Maybe renaming it to `check_commentable` would be better, then removing the `if not` check.

akx: Yeah, I think you're right. Will do that! :)

akx:
Squeezed in a migration that was missing, too.

Issue #126, opened 2016-01-08
Admin Labels box too narrow


Rikuoja:
Labels box when editing a hearing is so narrow that labels cannot be seen

Rikuoja:
Apparently the width is determined by the length of the available labels =)

PR #133, opened 2016-02-04
Fix labels field

tuomas777:
Set fixed width for labels and disabled the default help text which was giving wrong instructions in admin UI. 

Refs #126


akx:
LGTM, though FixedModelForm is a somewhat silly name imho :)

PR #135, opened 2016-02-08
Add preview url to Hearing admin detail page

tuomas777:
Based on UI_BASE_URL in settings file

Refs #134


akx:
Looks cromulent!

Issue #120, opened 2015-12-07
"Closure info" section type


akx:
A new section type that's hidden by default in the UI, but shown as the topmost one when the hearing is closed.

PR #137, opened 2016-02-12
Closure info section

tuomas777:
Add closure info section type

Don't copy closure info section to a new hearing draft

Refs #120


akx:
Very nice!

PR #139, opened 2016-02-16
Add map to hearing admin

tuomas777:
Removed Hearing's separate latitude and longitude fields, changed geojson to a django-geojson GeometryField and added Leaflet for the geojson field in the admin UI.

Contains also unrelated Pillow version update to fix https://github.com/python-pillow/Pillow/issues/1474

Refs https://github.com/City-of-Helsinki/kerrokantasi-ui/issues/116


akx:
This seems pretty reasonable. Good stuff!

PR #153, opened 2016-03-15
Add optional authorization code to comments

Rikuoja:


akx:
What does this mean/do? The comment has no refs to an issue describing the requirements for this change.

democracy/models/comment.py
```diff
@@ -11,6 +11,7 @@ class BaseComment(BaseModel):
     parent_model = None  # Required for factories and API
     title = models.CharField(verbose_name=_('title'), blank=True, max_length=255)
     content = models.TextField(verbose_name=_('content'))
+    authorization_code = models.CharField(verbose_name=_('authorization code'),  max_length=255, blank=True, null=True)
```
akx: CharFields generally shouldn't be nullable, since then you'll have two falsy values (`NULL` and `""`).

Is the field really required to be 255 characters long?

Rikuoja:
Ah, there is no issue for this yet. This is an ad hoc fix for the coming hearing, where we want to separate comments from people in residents' panel, where they have been given individual access codes. This cannot be done with user accounts, as the accounts of the people in the panel are not known.

democracy/models/comment.py
Rikuoja: I'll fix nullability. The code could indeed be shorter, e.g. 8 characters.

akx:
Fair enough, I suppose. :)

Though I would like to see an issue/spec/use-case before the actual implementation... :)

Rikuoja:
Yeah this is one of those uses that was communicated to me like one day before we are supposed to launch :D

PR #161, opened 2016-04-05
Include section comment count to hearing comment count

Rikuoja:


Rikuoja:
Fix #160

democracy/models/base.py
```diff
@@ -121,9 +121,16 @@ class Commentable(models.Model):
 
     def recache_n_comments(self):
         new_n_comments = self.comments.count()
+        # if commentable has sections, include them in the total comment count
+        if hasattr(self, 'sections'):
+            for section in self.sections.all():
```
akx: Use Django's `Sum` aggregate? (Or at the very least, use `.values_list()` to query only the `n_comments` field you need.

akx:
Missing tests.

akx:
CI failing still.

Please fold in the test changes to the commit making the corresponding changes in the actual code; otherwise using `git bisect` will be painful.

PR #168, opened 2016-04-12
Include plugin data in comment serialization, to obtain data for hearing visualization

Rikuoja:


akx:
I don't think this is a good idea, or at the very least it should be parametrized. Plugin data is potentially very large, and downloading all of it for every request (even those for the regular UI) is probably not actually desired.

akx:
Missing tests for both new features added.

democracy/views/comment.py
```diff
@@ -54,7 +62,13 @@ def get_serializer_context(self):
 
     def get_queryset(self):
         queryset = super().get_queryset()
-        return queryset.filter(**{queryset.model.parent_field: self.get_comment_parent_id()})
+        filtered_queryset = queryset.filter(**{queryset.model.parent_field: self.get_comment_parent_id()})
+        request = self.get_serializer_context().get('request', None)
+        if request:
+            auth_code = request.GET.get('auth_code', None)
```
akx: The request arg should probably be called `authorization_code` too.

How about using some sort of actual DRF filtering mechanism instead of hard-coding only one filter though?

democracy/tests/test_comment.py
```diff
@@ -40,6 +40,24 @@ def test_55_add_comment_to_hearing(john_doe, john_doe_api_client, default_hearin
...
+    assert comment_data["authorization_code"] ==\
+           Hearing.objects.get(pk=default_hearing.pk).comments.get(id=response_data["id"]).authorization_code
+
+
+@pytest.mark.django_db
+def test_list_comments_with_auth_code(api_client, default_hearing):
+    comment_data = get_comment_data(authorization_code="foo6")
+    url = get_hearing_detail_url(default_hearing.id, 'comments')
+    api_client.post(url, comment_data)
+    data = get_data_from_response(api_client.get(url+'?authorization_code=foo6'))
```
akx: `Client.get()` accepts the query string as a second parameter:

```python
api_client.get(url, {"authorization_code": "foo6"})
```

democracy/tests/test_comment.py
```diff
@@ -354,3 +372,25 @@ def test_add_plugin_data_to_comment(api_client, default_hearing, case):
...
+        section.plugin_identifier = "test_plugin"
+        section.save()
+        url = get_hearing_detail_url(default_hearing.id, 'sections/%s/comments' % section.id)
+        comment_data = get_comment_data(
+            content="",
+            plugin_data="foo6"
+        )
+        response = api_client.post(url, data=comment_data)
+        response_data = get_data_from_response(response, status_code=201)
+        comment_list = get_data_from_response(api_client.get(url+'?include=plugin_data'))
```
akx: Same GET complaint here

akx:
Can you add a test that ensures that plugin data is _not_ included when you don't pass in the `include=plugin_data` parameter?

akx:
Let's do this! Thanks.

PR #169, opened 2016-04-29
Add organization to hearing

Rikuoja:
Required by #138


democracy/models/base.py
```diff
@@ -116,6 +118,13 @@ class Meta:
         abstract = True
 
 
+class Organization(BaseModel):
+    name = models.CharField(verbose_name=_('name'), max_length=255)
+    admin_users = models.ManyToManyField(
+        User, blank=True, related_name='admin_organizations'
```
akx: Just use `settings.AUTH_USER_MODEL` directly here -- aliasing it to `User` makes the casual reader think it's not pluggable.

akx:
Missing tests.

What is the actual desired serialization format for `organization`? (Should it maybe be a `SlugRelatedField`?)

akx:
Also missing a migration.

democracy/migrations/0012_auto_20160502_1219.py
```diff
@@ -0,0 +1,41 @@
+# -*- coding: utf-8 -*-
```
akx: Please use the `-n` flag with `makemigrations` to specify the migration's name. (`auto` migration names are not very nice.)

democracy/migrations/0013_auto_20160502_1222.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
akx: Please use the `-n` flag with `makemigrations` to specify the migration's name. (`auto` migration names are not very nice.)

democracy/models/label.py
```diff
@@ -1,10 +1,10 @@
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
-from .base import BaseModel
+from .base import StringIdBaseModel
 
 
-class Label(BaseModel):
+class Label(StringIdBaseModel):
```
akx: This seems like an unrelated change to "Add organization to hearing"?

akx: And all in all, I'm not quite sure what this change is supposed to achieve. The commit message is quite opaque (and lacks any sort of issue ref).

Rikuoja: Agreed, it is only tangentially related. I created issue #171 for the change.

akx: Please add a reference to #171 in the commit message too.

akx:
@Rikuoja Rebase required.

akx:
Tossing the ball to @tuomas777 to review + merge.

democracy/models/base.py
```diff
@@ -115,6 +115,13 @@ class Meta:
         abstract = True
 
 
+class Organization(StringIdBaseModel):
+    name = models.CharField(verbose_name=_('name'), max_length=255, unique=True)
+    admin_users = models.ManyToManyField(
+        settings.AUTH_USER_MODEL, blank=True, related_name='admin_organizations'
+    )
```
tuomas777: IMO this model shouldn't be in `base.py` as it isn't used as a basis for anything

democracy/tests/test_hearing.py
```diff
@@ -214,6 +215,25 @@ def test_8_get_detail_labels(api_client):
 
 
 @pytest.mark.django_db
+def test_8_get_detail_organization(api_client):
+    hearing = Hearing()
+    hearing.save()
+
+    organization = Organization(name='The department for squirrel welfare')
+    organization.save()
```
tuomas777: You could use pytest fixture for hearing here, there is `default_hearing` for example. Also organization should be made as a fixture.

democracy/models/base.py
Rikuoja: Ah, didn't realize that was the point :)

democracy/migrations/0015_add_organization_model.py
```diff
@@ -0,0 +1,41 @@
...
+
+    operations = [
+        migrations.CreateModel(
+            name='Organization',
+            fields=[
+                ('created_at', models.DateTimeField(db_index=True, default=django.utils.timezone.now, editable=False, verbose_name='time of creation')),
+                ('modified_at', models.DateTimeField(default=django.utils.timezone.now, editable=False, verbose_name='time of last modification')),
+                ('published', models.BooleanField(db_index=True, default=True, verbose_name='public')),
+                ('deleted', models.BooleanField(db_index=True, default=False, editable=False, verbose_name='deleted')),
+                ('id', models.CharField(blank=True, help_text='You may leave this empty to automatically generate an ID', max_length=32, primary_key=True, serialize=False, verbose_name='identifier')),
```
tuomas777: Actually this is a bit wrong because `StringIdBaseModel` has changed, I *think* correct would be `('id', models.CharField(editable=False, max_length=32, primary_key=True, serialize=False, verbose_name='identifier'))`

tuomas777:
LGTM!

PR #174, opened 2016-05-11
Admin soft delete

### LINKED ISSUE #163: Cannot delete anything in admin (soft delete not implemented)

tuomas777:
Changed admin UI delete selected action to execute Hearing soft delete, and also added implementation for inlined Section and SectionImage soft deletion. Django-nested-admin needed to be updated to v3, which doesn't seem to cause any problems.

Closes #163


akx:
Looks almost as soft and fluffy as this bunny!

PR #177, opened 2016-05-13
Separate slug field for Hearings

### LINKED ISSUE #165: Separate slug from primary key to allow sensible, editable slugs

tuomas777:
Added a separate slug field for Hearings and changed ID to be non-editable. The slug field is used in admin UI the same way ID was before, it can be left empty to auto generate one etc. 

In the API, the slug can be used to access Hearings alongside ID, ie. both `/v1/hearing/<id>/` and `/v1/hearing/<slug>/` can be used for all operations.

Closes #165


democracy/locale/fi/LC_MESSAGES/django.po
```diff
@@ -196,6 +208,9 @@ msgstr "osat"
 msgid "service map URL"
 msgstr "palvelukartta-URL"
 
+msgid "slug"
+msgstr "slug"
```
akx: It's "lyhytnimi" in Finnish

democracy/tests/test_comment.py
```diff
@@ -21,17 +21,24 @@ def get_comment_data(**extra):
     }, **extra)
 
 
+@pytest.fixture(params=['id', 'slug'])
+def lookup_field(request):
```
akx: That's a meta fixture, yo.

democracy/views/section.py
```diff
@@ -46,7 +47,8 @@ class SectionViewSet(AdminsSeeUnpublishedMixin, viewsets.ReadOnlyModelViewSet):
     model = Section
 
     def get_queryset(self):
-        hearing = Hearing.objects.get(id=self.kwargs['hearing_pk'])
+        id_or_slug = self.kwargs['hearing_pk']
+        hearing = Hearing.objects.get(Q(pk=id_or_slug) | Q(slug=id_or_slug))
```
akx: Shouldn't this use `get_by_id_or_slug()`?

democracy/migrations/0012_add_hearing_slug.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+def forwards(apps, schema_editor):
+    """
+    Add slugs to existing Hearings.
+    """
+    hearing_model = apps.get_model('democracy', 'Hearing')
+    for hearing in hearing_model.objects.all():
+
+        # populate slugs from ids to avoid collisions which aren't easy to deal with here
```
akx: I guess this could just `get_random_string()` the slugs for existing hearings if that makes it any easier to deal with.

democracy/locale/fi/LC_MESSAGES/django.po
tuomas777: for real? :D well, actually was told to use "slugi" there, but apparently misspelled even that

democracy/views/section.py
tuomas777: yes

democracy/migrations/0012_add_hearing_slug.py
tuomas777: Using IDs is superior to random ones, because most of them should contain slugs, and they are hopefully unique also :) The point was just that normal slug from title cannot be used in this as that can lead to collisions.

PR #178, opened 2016-05-16
Add file uploading ability to CKEditor

### LINKED ISSUE #150: Allow CKEditor file uploads in hearing admin, to add links to files (pdfs etc.) relevant to the hearing

tuomas777:
By default django-ckeditor sets relative URLs for uploaded images
and for files selected using the editor's file browser, so those
views needed to be overidden to return absolute URLs.

Closes #150

### DIFF
--- a/democracy/views/upload.py
+++ b/democracy/views/upload.py
@@ -1,0 +1,89 @@
+import os
+
+from django.conf import settings
+from django.http import HttpResponse
+from django.shortcuts import render_to_response
+from django.template import RequestContext
+from django.views.decorators.csrf import csrf_exempt
+from ckeditor_uploader import image_processing
+from ckeditor_uploader import utils
+from ckeditor_uploader.views import get_files_browse_urls, ImageUploadView, SearchForm
+from django.utils.html import escape
+
+
+class AbsoluteUrlImageUploadView(ImageUploadView):
+    http_method_names = ['post']
+
+    def post(self, request, **kwargs):
+        """
+        Uploads a file and send back its URL to CKEditor.
+
+        Exactly the same as in django-ckeditor 5.0.3 except that creates
+        absolute URLs instead of relative ones.
+        """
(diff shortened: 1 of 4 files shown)


democracy/views/upload.py
```diff
@@ -0,0 +1,86 @@
...
+            except utils.NotAnImageException:
+                return HttpResponse("""
+                    <script type='text/javascript'>
+                    window.parent.CKEDITOR.tools.callFunction({0}, '', 'Invalid file type.');
+                    </script>""".format(ck_func_num))
+
+        saved_path = self._save_file(request, uploaded_file)
+        self._create_thumbnail_if_needed(backend, saved_path)
+        url = utils.get_media_url(saved_path)
+        url = request.build_absolute_uri(url)
```
JsseL: So is this the only customization to this method?

JsseL:
Some comments might clarify which lines were your customations. I'll assume the original code is tested and therefore only the customized part is interesting at this point.

Otherwise looks pretty good.

tuomas777:
Added comments to the customizations.

tuomas777:
Doesn't have tests, but as is the case with many other admin UI customizations, IMHO needed work / reward ratio isn't there, especially on this one.

JsseL:
Exactly, most of the code is already tested by the upstream dependency and now that I see you added only few robust looking lines I definitely agree with you. :shipit: good!

akx:
This looks fine to me - feel free to merge if you want.

However, would it possibly be cleaner to postprocess the HTML from CKEditor (using `lxml.html`?), find all `img`s (and other embeds if there are any) and ensure their `src`s and `href`s and whatnot are absolute?

tuomas777:
Yeah postprocessing the HTML would have been another option, but as this one is ready and IMO pretty ok also I don't think it warrants a change. So I'll go ahead and merge, the postprocessing can be implemented later if needed

PR #181, opened 2016-05-24
Flat comment and image endpoints

tuomas777:
Converted `HearingComment`s and `HearingImage`s to `SectionComment`s and `SectionImage`s of introduction section and removed the old models.

Added new root level endpoints `/v1/comment/` and `/v1/image/` which should behave the same way as the old comment and image endpoints. Some tests are used for both the new and the old endpoints.

Added also `main_image` field to `Hearing` endpoint to make the hearing's introduction section first image available also in `Hearing' list.

ATTENTION: because `/v1/hearing/<id>/comments/` and `/v1/hearing/<id>/images/` are removed, the frontend requires also changes.


democracy/migrations/0016_merge_comments.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('democracy', '0015_add_organization_model'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, backwards),
```
akx: Use `migrations.RunPython.noop` as backwards (saves you from writing the same fn yourself :) )

democracy/migrations/0016_merge_comments.py
```diff
@@ -0,0 +1,56 @@
...
+    hearing_comments = apps.get_model('democracy', 'HearingComment').objects.all()
+    section_comment_model = apps.get_model('democracy', 'SectionComment')
+
+    for hearing_comment in hearing_comments:
+        data = {field: getattr(hearing_comment, field) for field in ('content', 'author_name', 'n_votes', 'created_by',
+                                                                     'created_at', 'modified_at', 'modified_by')}
+        section = hearing_comment.hearing.sections.get(type__identifier='introduction')
+        data['section'] = section
+
+        section_comment_model.objects.create(**data)
```
akx: Will this deal properly with migrating comment revision history?

democracy/migrations/0016_merge_comments.py
```diff
@@ -0,0 +1,56 @@
...
+
+def forwards(apps, schema_editor):
+    """
+    Convert Hearing comments to introduction Section comments
+    """
+    hearing_comments = apps.get_model('democracy', 'HearingComment').objects.all()
+    section_comment_model = apps.get_model('democracy', 'SectionComment')
+
+    for hearing_comment in hearing_comments:
+        data = {field: getattr(hearing_comment, field) for field in ('content', 'author_name', 'n_votes', 'created_by',
```
akx: Use `django.forms.models.model_to_dict()` :)

democracy/migrations/0017_merge_images.py
```diff
@@ -0,0 +1,53 @@
...
+
+def forwards(apps, schema_editor):
+    """
+    Convert Hearing images to introduction Section images
+    """
+    hearing_images = apps.get_model('democracy', 'HearingImage').objects.all()
+    section_image_model = apps.get_model('democracy', 'SectionImage')
+
+    for hearing_image in hearing_images:
+        data = {field: getattr(hearing_image, field) for field in ('title', 'caption', 'height', 'width', 'image',
```
akx: Use `django.forms.models.model_to_dict()`

democracy/migrations/0017_merge_images.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('democracy', '0016_merge_comments'),
+    ]
+
+    operations = [
+        migrations.RunPython(forwards, backwards),
```
akx: Same backwards nit here

democracy/models/base.py
```diff
@@ -124,10 +124,6 @@ class Commentable(models.Model):
 
     def recache_n_comments(self):
         new_n_comments = self.comments.count()
-        # if commentable has sections, include them in the total comment count
```
akx: Are you sure this is the right change? I imagined Hearings should tally the total number of comments in their Sections, so you could for instance sort by the most commentedest hearing.

democracy/models/hearing.py
```diff
@@ -96,23 +94,16 @@ def save(self, *args, **kwargs):
...
-    parent_field = "hearing"
-    parent_model = Hearing
-    hearing = models.ForeignKey(Hearing, related_name="comments")
-
-    class Meta:
-        verbose_name = _('hearing comment')
-        verbose_name_plural = _('hearing comments')
+    def recache_n_comments(self):
+        new_n_comments = 0
+        if self.sections.all():
```
akx: Unnecessary `if` (that causes a database query).

Instead just 

```python
new_n_comments = self.sections.all().aggregate(Sum('n_comments')).get('n_comments__sum') or 0
```

democracy/tests/test_comment.py
```diff
@@ -21,162 +21,62 @@ def get_comment_data(**extra):
...
-
-
-@pytest.mark.django_db
-def test_54_list_all_comments_added_to_hearing_check_all_properties(api_client, default_hearing):
-    # list all comments
-    response = api_client.get(get_hearing_detail_url(default_hearing.id, 'comments'))
-    data = get_data_from_response(response)
-    # get first returned comment
-    comment = data[0]
+@pytest.fixture(params=[0, 1, 2])
```
akx: Maybe give the params descriptive names?

democracy/tests/test_comment.py
```diff
@@ -426,3 +314,41 @@ def test_get_plugin_data_for_comment(api_client, default_hearing):
...
+    comment = default_hearing.sections.first().comments.first()
+    url = '%s%s/' % (url, comment.id)
+
+    response = api_client.put(url, data)
+    assert response.status_code == 400
+    assert 'section' in response.data
+
+
+@pytest.mark.django_db
+def test_root_endpoint_filters(api_client, default_hearing, random_hearing):
```
akx: Unnecessary `random_hearing` :)

democracy/tests/test_comment_vote.py
```diff
@@ -27,18 +27,9 @@ def get_section_comment_vote_url(hearing_id, section_id, comment_id):
...
-def get_hearing_comment_vote_url(hearing_id, comment_id):
-    # /v1/hearings/<hearingID>/comments/<commentID>/votes/
-    return get_hearing_detail_url(hearing_id, 'comments/%s/vote' % comment_id)
-
-
-def get_hearing_comment_unvote_url(hearing_id, comment_id):
-    # /v1/hearings/<hearingID>/comments/<commentID>/unvotes/
-    return get_hearing_detail_url(hearing_id, 'comments/%s/unvote' % comment_id)
+def get_section_comment_unvote_url(hearing_id, section_id, comment_id):
+    # /v1/hearings/<hearingID>/sections/<sectionID>/comments/<commentID>/unvotes/
```
akx: `/unvote/`

democracy/tests/test_images.py
```diff
@@ -1,12 +1,12 @@
...
 
-from democracy.tests.test_hearing import list_endpoint
-from democracy.tests.utils import IMAGES, get_data_from_response, get_hearing_detail_url
+from democracy.tests.utils import IMAGES, create_default_images, get_data_from_response, get_hearing_detail_url
 
 
-def check_entity_images(entity):
-    assert 'images' in entity
-    image_list = entity['images']
+def check_entity_images(entity, images_field=True):
```
akx: Maybe `require_images` instead of `images_field`?

akx: Would just be better to directly pass `entity['images']` instead of having this peel it off?

democracy/views/hearing.py
```diff
@@ -62,23 +44,40 @@ def get_sections(self, hearing):
         serializer.bind('sections', self)  # this is needed to get context in the serializer
         return serializer.to_representation(queryset)
 
+    def get_main_image(self, hearing):
```
akx: The performance (or possible lack thereof) of this method slightly worries me. Looks like it might be doing a whole lot of queries and getting lotsa objects just to maybe return one...

```python

main_image_qs = SectionImage.objects.filter(
  section__hearing=hearing,
  section__type__identifier='introduction',
)

if not self.context['request'].user.is_superuser:
  main_image_qs = main_image_qs.filter(published=True)

main_image = main_image_qs.first()
```
would be only one query.

democracy/views/section.py
```diff
@@ -52,3 +54,34 @@ def get_queryset(self):
         if not hearing.closed:
             queryset = queryset.exclude(type__identifier=InitialSectionType.CLOSURE_INFO)
         return queryset
+
+
+class RootSectionImageSerializer(SectionImageSerializer):
```
akx: A docstring to explain the raison d'être for this serializer would be nice.

democracy/views/section.py
```diff
@@ -52,3 +54,34 @@ def get_queryset(self):
...
+    hearing = serializers.SerializerMethodField()
+
+    class Meta(SectionImageSerializer.Meta):
+        fields = SectionImageSerializer.Meta.fields + ['section', 'hearing']
+
+    def get_hearing(self, section_image):
+        return section_image.section.hearing.id
```
akx: Yigh, at least use `hearing_id` (and it would be a darn good idea to ensure `.section` is either cached in or `.select_related()`ed.

democracy/views/section.py
```diff
@@ -52,3 +54,34 @@ def get_queryset(self):
         if not hearing.closed:
             queryset = queryset.exclude(type__identifier=InitialSectionType.CLOSURE_INFO)
         return queryset
+
+
+class RootSectionImageSerializer(SectionImageSerializer):
+    hearing = serializers.SerializerMethodField()
```
akx: Hold on, if you only need the hearing ID as the field `hearing`, do
```python
hearing = serializers.CharField(read_only=True, source='hearing_id')
```

and be done with it.

democracy/views/section.py
```diff
@@ -52,3 +54,34 @@ def get_queryset(self):
...
+
+    class Meta(SectionImageSerializer.Meta):
+        fields = SectionImageSerializer.Meta.fields + ['section', 'hearing']
```
akx: This list declares `section` but the serializer doesn't? Is it in a superclass?

democracy/views/section_comment.py
```diff
@@ -56,3 +58,74 @@ class SectionCommentViewSet(BaseCommentViewSet):
     model = SectionComment
     serializer_class = SectionCommentSerializer
     create_serializer_class = SectionCommentCreateSerializer
+
+
+class RootSectionCommentSerializer(SectionCommentSerializer):
+    hearing = serializers.SerializerMethodField()
```
akx: Same complaint about how to serialize the hearing id.

democracy/views/section_comment.py
```diff
@@ -56,3 +58,74 @@ class SectionCommentViewSet(BaseCommentViewSet):
...
+        return serializer_class(*args, **kwargs)
+
+    def get_comment_parent_id(self):
+        parent_field = self.get_serializer_class().Meta.model.parent_field
+        try:
+            return self.request.data[parent_field]
+        except KeyError:
+            return None
+
+    def get_comment_parent(self):
```
akx: I think this could be simplified now that we know Comments' parents are always Sections..?

democracy/views/section_comment.py
```diff
@@ -56,3 +58,74 @@ class SectionCommentViewSet(BaseCommentViewSet):
...
+
+class CommentFilter(filters.FilterSet):
+    hearing = django_filters.CharFilter(name='section__hearing__id')
+
+    class Meta:
+        model = SectionComment
+        fields = ['authorization_code', 'section', 'hearing']
+
+
+# root level (Section)Comment endpoint
```
akx: Well it always is SectionComment, isn't it? :)

akx:
First pass of review done, passing ball back to @tuomas777 for implementation.

democracy/migrations/0016_merge_comments.py
tuomas777: Probably not, didn't know we had any relevant comment history yet. Do we?

democracy/migrations/0016_merge_comments.py
tuomas777: That one doesn't include non editable fields? `created_at`, `created_by` etc

democracy/tests/test_comment.py
tuomas777: Actually it isn't, it holds comments that should not be included if the filters work correctly.

democracy/tests/test_images.py
tuomas777: Well yeah, but it is used by other tests and in general if possible I try not to change old tests (especially other people written ones). Of course this one would have been a simple change and this solution is a bit awkward, but meh, it's only tests :)

democracy/views/hearing.py
tuomas777: Yeah good points about performance, wasn't really aiming for performance, but especially this one was way too naive. However I think this solution of yours isn't quite there, because I think we should be prepared for multiple introduction images, in which case it would possible work incorrectly.

akx: Don't `SectionImage`s have `ordering = ('ordering',)` or similar in their Metas? That'd mean `.first()` implicitly grabs the first-ordered image from the intro section...

democracy/tests/test_comment.py
akx: Ah, okay!

democracy/migrations/0016_merge_comments.py
akx: Ah, you're right. Sorry!

democracy/migrations/0016_merge_comments.py
akx: Dunno?

democracy/views/section.py
tuomas777: The model `SectionImage` has field `section`, or what do you mean?

democracy/views/section_comment.py
tuomas777: Well the point of the comment was that this is a comment endpoint, for which `SectionComment` objects are used. Clearly the comment was a fail as the first reader got it wrong :D

democracy/views/hearing.py
tuomas777: Yes, but I'm thinking of a situation where there are two images, first one non public and second one public. In that case the returned main image would be a different image between normal user and superuser, which doesn't seem good.

democracy/views/section_comment.py
tuomas777: Yeah and it could be simplified probably in other ways as well, but I think that can wait until the final API structure has been decided to avoid making changes back and forth

tuomas777:
Implemented most of the proposed changes and made comments on the rest.

akx:
LGTM.

PR #188, opened 2016-06-03
Flat endpoint embetterments and stuff

tuomas777:
- Added `Section` root level endpoint

- Fixed root level endpoints to exclude objects if parent `Hearing` if excluded

- Removed `commenting` and `abstract` from `Hearing`. During the migration introduction section is populated by values of those.

- Improved error handling in image and comment merge migrations


akx:
What's the reasoning behind removing the hearing abstract? I think it was/is useful for showing a little blurb about the hearing in listings, where you don't really care about the sections at all.

tuomas777:
Well, now that whole hearing related stuff goes into the intro section, there isn't a need for both hearing abstract and intro section abstract. Maybe another option would have been to keep abstract in hearing and then just keep abstract in intro section always empty.

akx:
Well, lgtm, I guess -- though I kinda think this should've been two PRs (one for the section endpoints, another for the hearing changes).

tuomas777:
Yeah true, could have done two PRs.

PR #189, opened 2016-06-03
Rename section type introduction to main

tuomas777:
Because the section has changed to include stuff related to the whole `Hearing`, the old name wasn't describing it well anymore.


democracy/migrations/0009_section_type_model.py
```diff
@@ -7,12 +7,48 @@
 from django.db import migrations, models
 import django.db.models.deletion
 import django.utils.timezone
-from democracy.models.initial_data import create_initial_section_types
+
+
+class InitialSectionType:
```
akx: Hmm. Will it actually matter if this migration creates the section types with new data?

democracy/migrations/0020_rename_intro_section_to_main.py
```diff
@@ -0,0 +1,48 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations
+
+
+def forwards(apps, schema_editor):
+    """
+    Rename introduction section type to main
+    """
+    section_type_model = apps.get_model('democracy', 'SectionType')
```
akx: These migrations really could use [queryset updates](https://docs.djangoproject.com/en/1.9/ref/models/querysets/#update) instead:

```python
section_type_model.objects.filter(identifier='introduction').update(identifier=..., name_singular=...)
```

democracy/migrations/0009_section_type_model.py
tuomas777: There are later migrations that assume there exists a type `introduction`. While that should not matter if there is no actual data in the db, I still think this is more safer way without any drawbacks.

PR #190, opened 2016-06-03
Switch to helusers admin

Rikuoja:


requirements.txt
```diff
@@ -1,3 +1,4 @@
+git+https://github.com/City-of-Helsinki/django-helusers.git#egg=django-helusers
```
akx: This will always get the `master` version of helusers, with no pinning whatsoever. Is that a good idea?

akx:
This seems to do a whole lot of changes, and the commit message does not explain why, nor does it have refs to an issue or anything. :(

akx:
Tossing ball back to @Rikuoja.

requirements.txt
```diff
@@ -19,4 +19,5 @@ django-nested-admin==3.0.2
 django-geojson==2.9.0
 django-leaflet==0.18.0
 django-ckeditor==5.0.3
-django-autoslug==1.9.3
\ No newline at end of file
+django-autoslug==1.9.3
+django-allauth
```
JsseL: Might be good idea to pin the `django-allauth` too. Pinned versions makes things more robust.

PR #193, opened 2016-06-07
Show not yet opened Hearings if the user is super or has the preview code

### LINKED ISSUE #191: Return future hearings if logged in or provided with preview link
Enhancement on #170. Current implementation makes future hearings completely disappear and not previewable.

tuomas777:
Closes #191

### DIFF
--- a/democracy/tests/test_hearing.py
+++ b/democracy/tests/test_hearing.py
@@ -266,6 +266,7 @@ def test_can_see_unpublished_with_preview_code(api_client):
     hearings = create_hearings(1)
     unpublished_hearing = hearings[0]
     unpublished_hearing.published = False
+    unpublished_hearing.open = now() - datetime.timedelta(hours=1)
     unpublished_hearing.save()
     get_data_from_response(api_client.get(get_detail_url(unpublished_hearing.id)), status_code=404)
     preview_url = "{}?preview={}".format(get_detail_url(unpublished_hearing.id), unpublished_hearing.preview_code)
@@ -325,18 +326,26 @@ def test_hearing_copy(default_hearing, random_label):
     assert not new_hearing.sections.filter(type__identifier=InitialSectionType.CLOSURE_INFO).exists()
 
 
+@pytest.mark.parametrize('client, expected', [
+    ('api_client', False),
+    ('jane_doe_api_client', False),
+    ('admin_api_client', True)
+])
 @pytest.mark.django_db
-def test_hearing_open_at_filtering(api_client, default_hearing):
+def test_hearing_open_at_filtering(default_hearing, request, client, expected):
+    api_client = request.getfuncargvalue(client)
+
     default_hearing.open_at = now() + datetime.timedelta(hours=1)
     default_hearing.save(update_fields=('open_at',))
 
     response = api_client.get(list_endpoint)
     data = get_data_from_response(response)
     ids = [hearing['id'] for hearing in data]
-    assert default_hearing.id not in ids
+    assert bool(default_hearing.id in ids) == expected
 
+    expected_code = (200 if expected else 404)
     response = api_client.get(get_hearing_detail_url(default_hearing.id))
-    assert response.status_code == 404
+    assert response.status_code == expected_code
 
 
 @pytest.mark.django_db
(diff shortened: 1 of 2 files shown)


democracy/tests/test_hearing.py
```diff
@@ -325,18 +326,27 @@ def test_hearing_copy(default_hearing, random_label):
...
+    api_client = request.getfuncargvalue(client)
+
     default_hearing.open_at = now() + datetime.timedelta(hours=1)
     default_hearing.save(update_fields=('open_at',))
 
     response = api_client.get(list_endpoint)
     data = get_data_from_response(response)
     ids = [hearing['id'] for hearing in data]
-    assert default_hearing.id not in ids
+    assert bool(default_hearing.id in ids) is expected
```
akx: use `==`, not `is` (though it doesn't actually matter in this particular case)

democracy/views/hearing.py
```diff
@@ -134,32 +133,39 @@ def get_queryset(self):
                 to_attr='main_section_list'
             )
         )
+
+        if not self.request.user.is_superuser:
```
akx: Add a comment describing this filter. Something like

```python
# Unless the user is a superuser, only show open hearings.
```

or similar.

democracy/views/hearing.py
```diff
@@ -134,32 +133,39 @@ def get_queryset(self):
...
-        is_superuser = user and user.is_authenticated() and user.is_superuser
-
+        is_superuser = self.request.user.is_superuser
+        preview_code = None
         if not obj.published and not is_superuser:
             preview_code = self.request.query_params.get('preview')
             if not preview_code or preview_code != obj.preview_code:
                 raise NotFound()
 
+        if not (preview_code or is_superuser or obj.open_at <= now()):
```
akx: Why is the open_at check required here? Won't the queryset already have such objects filtered out?

democracy/views/hearing.py
tuomas777: In general I could do more commenting I give you that, but I don't really understand this one, the two lines of code are saying just the same thing? :D

democracy/views/hearing.py
tuomas777: No those are not yet filtered out, and they cannot be, because we want those to be shown if the user is a superuser or has the preview code

tuomas777:
implemented suggested changes

democracy/tests/test_hearing.py
```diff
@@ -325,18 +326,26 @@ def test_hearing_copy(default_hearing, random_label):
...
 
+@pytest.mark.parametrize('client, expected', [
+    ('api_client', False),
+    ('jane_doe_api_client', False),
+    ('admin_api_client', True)
+])
 @pytest.mark.django_db
-def test_hearing_open_at_filtering(api_client, default_hearing):
+def test_hearing_open_at_filtering(api_client, default_hearing, request, client, expected):
+    api_client = request.getfuncargvalue(client)
```
akx: You're reassigning the param here -- maybe remove `api_client` from the param list?

democracy/tests/test_hearing.py
```diff
@@ -325,18 +326,26 @@ def test_hearing_copy(default_hearing, random_label):
...
     default_hearing.open_at = now() + datetime.timedelta(hours=1)
     default_hearing.save(update_fields=('open_at',))
 
     response = api_client.get(list_endpoint)
     data = get_data_from_response(response)
     ids = [hearing['id'] for hearing in data]
-    assert default_hearing.id not in ids
+    assert bool(default_hearing.id in ids) == expected
 
+    expected_code = 200 if expected else 404
```
akx: Parens around the rvalue, please :)

democracy/views/hearing.py
```diff
@@ -134,32 +133,41 @@ def get_queryset(self):
...
-
+        is_superuser = self.request.user.is_superuser
+        preview_code = None
         if not obj.published and not is_superuser:
             preview_code = self.request.query_params.get('preview')
             if not preview_code or preview_code != obj.preview_code:
                 raise NotFound()
 
+        # require preview_code or superuser status to show a not yet opened hearing
+        if not (preview_code or is_superuser or obj.open_at <= now()):
```
akx: Won't 162..165 already deal with previewcodeness? Does this need to bother about it?

tuomas777: 162-165 checks if preview code exists and is valid, this line excludes not yet opened hearings in case there is no preview_code or superuser. It might be possible to combine this into 162-165 ifs somehow, but I think this approach is also ok.

PR #194, opened 2016-06-08
Admin UI improvements

tuomas777:
- Validate number of main and closure info sections

- Don't show public flag for sections or labels


akx:
This validation sounds like it's something a Hearing model should do instead of being a responsibility of the admin UI? Or am I wrong in assuming that?

tuomas777:
Well, it depends on how Hearing and it's Sections creation via API will be handled in the future. Also I think it is hard or even impossible to validate Sections in Hearing model in all cases.

tuomas777:
updated the second commit, fixed labels there as well

PR #198, opened 2016-06-20
Allow setting comment author name via API

tuomas777:
Refs City-of-Helsinki/kerrokantasi-ui#208


democracy/views/comment.py
```diff
@@ -91,8 +91,13 @@ def create(self, request, *args, **kwargs):
         if resp:
             return resp
 
+        # logged in users cannot set author_name
+        data = request.data.copy()
+        if request.user.is_authenticated():
+            data.pop('author_name', None)
```
akx: Maybe this should raise some sort of error instead of quietly doing things the user doesn't expect.

Rikuoja:
Requires #206 implemented.

tuomas777:
Added error message

PR #200, opened 2016-06-27
Prevent importer from deleting main section just created when patching

Rikuoja:


Rikuoja:
Any ideas @tuomas777 how to fix the importer while not exceeding flake8 complexity criteria? The admittedly stupid earlier implementation was in place exactly because the build did not go through otherwise.

akx:
@Rikuoja You can reduce the cyclomatic complexity of a function by refactoring it into several smaller functions.

Rikuoja:
@akx That is evident, and what I tried to do in the earlier implementation. The call was for any sensible actual implementations of said idea in this particular question.

akx:
@Rikuoja Simply move lines 169..176 in the postimage (I mean the new code) to, say, `clean_hearing_for_patching(hearing)`?

akx:
Likewise, the section number compacting code could maybe live as `Hearing.compact_section_ordering()`; you might want to call it from the admin or wherever when sections change.

Rikuoja:
Yay!

democracy/importing/json_importer.py
```diff
@@ -175,25 +166,34 @@ def import_hearing(hearing_datum, force=False, patch=False):
...
+    import_images(main_section, hearing_datum)
+    import_sections(hearing, hearing_datum, force)
+    compact_section_ordering(hearing)
+    if hearing_datum.keys():  # pragma: no cover
+        log.warn("These keys were not handled while importing %s: %s", hearing, hearing_datum.keys())
+    return hearing
 
-    # Compact section ordering...
+
+def clean_hearing_for_patching(hearing):
```
akx: So the images and comments are no longer cleaned at all?

Rikuoja: Yep. The lines were removed as actually they no longer are related to the Hearing in question, so trying to clean them would result in an error.

PR #202, opened 2016-06-28
Add KSV plugin support

Rikuoja:


akx:
I don't know what this is, but I AM OKAY WITH IT.

PR #208, opened 2016-07-06
Do not return user data within comments

### LINKED ISSUE #206: Do not return user UUID with "comment" endpoint
Only return Boolean is_registered, and user nickname.

tuomas777:
Added is_registered field which determines whether the author is
registered or not.

Closes #206


democracy/views/comment.py
```diff
@@ -24,6 +25,9 @@ def to_representation(self, instance):
                 r['plugin_data'] = instance.plugin_data
         return r
 
+    def get_is_registered(self, obj):
+        return obj.created_by is not None
```
akx: Use the FK backing field `created_by_id` to avoid loading a foreign object.

akx:
Fair enough, though the original issue #206 has absolutely no reasoning behind this change...

PR #209, opened 2016-07-06
Implement management command for removing duplicate comments

juyrjola:


democracy/management/commands/democracy_remove_dupes.py
```diff
@@ -0,0 +1,41 @@
...
+                    print("\tdifferent parent %s" % other)
+                    continue
+                if other.plugin_data != first.plugin_data:
+                    print("\tplugin data differs %s" % other)
+                    continue
+                print("\tbye bye %s" % other)
+                other.soft_delete()
+
+    def handle(self, *args, **options):
+        if not options.pop("nothing_can_go_wrong", False):
```
akx: Wouldn't it be better that this mgmt cmd printed what it was about to do, but only actually called `soft_delete` when you passed in the `--i-am-not-a-dummy` flag?

democracy/management/commands/democracy_remove_dupes.py
```diff
@@ -0,0 +1,41 @@
...
+from democracy.models import SectionComment
+
+
+class Command(BaseCommand):
+    option_list = BaseCommand.option_list + (
+        make_option("--yes-i-know-what-im-doing", dest="nothing_can_go_wrong", action="store_true"),
+    )
+
+    def _remove_dupes(self, klass):
+        potential_dupes = klass.objects.values('content',).exclude(content='')\
```
Rikuoja: Here we cannot actually exclude empty content. Lots of identical plugin_data have been posted with empty content, and duplicate plugin_data should be removed as well.

PR #218, opened 2016-09-06
Update default configuration to allow CORS accesses to URLs

vikoivun:
It is unclear whether we need these restrictions at all, but this the minimal change to allow the API to work by default.


akx:
@vikoivun It'd be awfully nice to not allow CORS accesses to, say the admin, so yeah, I'd say it's good to restrict it some.

vikoivun:
Point. Really would be nice, if admin&such were in another URL space.

PR #227, opened 2016-10-04
Add label and contact person endpoints

tuomas777:
Implemented root level read-only API endpoints for `Label` and `ContactPerson` models. 

Fields: 
Label: `id` and `label`
ContactPerson : `id`, `title`, `phone`, `email`, `name` and `organization` (`Organization.name`)

Also added a common default limit pagination class.


democracy/tests/test_contact_person.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+@pytest.mark.django_db
+def test_get_contact_person_list_check_fields(api_client, contact_person):
+    data = get_data_from_response(api_client.get('/v1/contact_person/'))
+    assert len(data['results']) == 1
+
+    contact_person_data = data['results'][0]
+    assert set(contact_person_data.keys()) == {'id', 'title', 'phone', 'email', 'name', 'organization'}
+    assert_common_keys_equal(contact_person_data, {
```
JsseL: This sounds like if there is no common keys this would still pass. That's most likely not what it does though is it?

tuomas777: Hmm, not sure what you mean, that is what it does :) And the `assert` on line 13 makes sure there are common keys?

JsseL: Ah the line 13 assures this makes sense. My bad. 👍

PR #229, opened 2016-10-24
Add necessary fields for a fullscreen plugin with customizable source code

Rikuoja:


Rikuoja:
This is required for front-end issue #251.

democracy/models/section.py
```diff
@@ -49,6 +49,8 @@ class Section(Commentable, StringIdBaseModel):
     content = models.TextField(verbose_name=_('content'), blank=True)
     plugin_identifier = models.CharField(verbose_name=_('plugin identifier'), blank=True, max_length=255)
     plugin_data = models.TextField(verbose_name=_('plugin data'), blank=True)
+    plugin_source = models.URLField(verbose_name=_('plugin source code'), blank=True)
```
akx: Are you sure these need to be per-section? The original idea was that they could be mapped via the `plugin identifier`s.

I.e.

```python
DEMOCRACY_PLUGINS = {
  "mapdon-hkr": "democracy.plugins.MapdonHkrPlugin",
}
```

```python
class MapdonHkrPlugin(Plugin):
    source = "https://example.com/js"
    fullscreen = True
```

akx:
The commit messages have no reference to an GH issue that requires this functionality :(

Rikuoja:
Yeah, the issue is a frontend issue, don't want to close it with backend commit.

akx:
At least add a cross-project ref (doesn't have to be "closes"):

> Refs City-of-Helsinki/kerrokantasi-ui#251

democracy/models/section.py
akx: Right now the base plugin class doesn't give a damn about the format of the data; the idea was that should the data need some validation, that plugin class could do just that. (Or, maybe, stash the data in a new model or something? Anyway.)

To avoid hard-coding things, maybe turn all this into a `Plugin` model:

* `identifier` (not strictly necessary, I guess)
* `iframe_url`
* `fullscreen`

and add a foreign key to it in Section?

Rikuoja: Hm, the thing is the same source code allows rendering both fullscreen and not. Therefore, that is a property of the desired section, and might be triggered in the UI.

Hence, the only field needed for the model would be the iframe_url. Do we want a model for that?

akx: Fair enough. At least rephrase `plugin_source` as `plugin_iframe_url` then?

Also, I suppose `plugin_identifier` (and the `Plugin` class structure) will have become obsolete by this change, so consider removing them?

Rikuoja: Hm, I would keep the plugin_identifier since we probably will have non-map-questionnaires that want completely different data in the future. This way, we can tell the frontend to use a different React component when there is a need for that.

Rikuoja: Actually, I should rename `plugin_identifier`to `plugin_type`since that is what it means.

Rikuoja: ...or I would if the API and frontend didn't rely on it being called `plugin_identifier` :/

Rikuoja: But yeah @akx now `plugin_iframe_url` is there at least.

kerrokantasi/settings/base.py
```diff
@@ -114,7 +114,8 @@
 DEMOCRACY_PLUGINS = {
     "mapdon-hkr": "democracy.plugins.Plugin",  # TODO: Create an actual class for this once we know the data format
     "mapdon-ksv": "democracy.plugins.Plugin",
-    "mapdon-ksv-visualize": "democracy.plugins.Plugin"
+    "mapdon-ksv-visualize": "democracy.plugins.Plugin",
+    "map-questionnaire": "democracy.plugins.Plugin"
```
akx: Trailing comma, please.

democracy/views/section.py
```diff
@@ -32,7 +32,8 @@ class Meta:
         fields = [
             'id', 'type', 'commenting', 'published',
             'title', 'abstract', 'content', 'created_at', 'created_by', 'images', 'n_comments',
-            'plugin_identifier', 'plugin_data', 'type_name_singular', 'type_name_plural'
+            'plugin_identifier', 'plugin_data', 'type_name_singular', 'type_name_plural',
+            'plugin_iframe_url', 'plugin_fullscreen'
```
akx: Trailing comma, please.

akx: Also group all the plugin stuff on one line, please?

PR #232, opened 2016-10-26
Allow anonymous voting if desired

### LINKED ISSUE #231: Allow unregistered voting on section comments if desired

Rikuoja:
Closes #231


democracy/models/comment.py
```diff
@@ -24,11 +24,10 @@ class BaseComment(BaseModel):
...
     )
-    n_legacy_votes = models.IntegerField(
-        verbose_name=_('legacy vote count'),
-        help_text=_('number of votes imported from legacy system'),
-        default=0,
-        editable=False
+    n_unregistered_votes = models.IntegerField(
+        verbose_name=_('unregistered vote count'),
+        help_text=_('number of unregistered votes'),
+        default=0
```
akx: Why was the editable flag dropped? I don't think this should be editable by humans. ([The `editable` flag does not make a field read-only.](https://docs.djangoproject.com/en/1.10/ref/models/fields/#editable))

democracy/views/comment.py
```diff
@@ -137,22 +148,27 @@ def perform_update(self, serializer):
...
-        comment.recache_n_votes()
-        # return success
-        return response.Response({'status': 'Vote has been added'}, status=status.HTTP_201_CREATED)
+        if not request.user.is_authenticated():
+            # If the check went through, anonymous voting is allowed
+            comment.n_unregistered_votes += 1
+            comment.recache_n_votes()
+            return response.Response({'status': 'Vote has been counted'}, status=status.HTTP_200_OK)
+        else:
+            # Check if user voted already. If yes, return 400.
```
akx: The comment lies. (304 is returned, not 400.)

democracy/views/comment.py
```diff
@@ -137,22 +148,27 @@ def perform_update(self, serializer):
...
-        # update number of votes
-        comment.recache_n_votes()
-        # return success
-        return response.Response({'status': 'Vote has been added'}, status=status.HTTP_201_CREATED)
+        if not request.user.is_authenticated():
+            # If the check went through, anonymous voting is allowed
+            comment.n_unregistered_votes += 1
+            comment.recache_n_votes()
+            return response.Response({'status': 'Vote has been counted'}, status=status.HTTP_200_OK)
+        else:
```
akx: Unnecessary `else:`; the previous branch has a `return`. (Removing the `else:` will also have the pleasant side effect of shortening the diff, as the rest of the function remains untouched.)

democracy/models/comment.py
Rikuoja: Ah, I did indeed mistake one for the other.

Rikuoja:
@akx Fixed

PR #242, opened 2016-12-02
Allow to POST, PUT and PATCH Hearings

stephawe:
Add serializers for Hearing create/update
Restrict PUT/PATCH to users of the same organization
Remove the unused section field "created_by" from section serializer
Allow user to GET unpublished hearings from their own organization
Add Tests

### DIFF
--- a/democracy/tests/test_hearing.py
+++ b/democracy/tests/test_hearing.py
@@ -118,6 +118,12 @@ def _update_hearing_data(data):
     data['sections'][1]['images'][0]['caption'] = 'New image caption'
 
 
+@pytest.fixture
+def unpublished_hearing_json(valid_hearing_json):
+    valid_hearing_json.update({'published': False})
+    return valid_hearing_json
+
+
 def get_detail_url(id):
     return '%s%s/' % (endpoint, id)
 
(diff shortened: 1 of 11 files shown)


democracy/tests/test_hearing.py
```diff
@@ -6,17 +6,124 @@
...
+                "plugin_data": "",
+                "type_name_singular": "osa-alue",
+                "type_name_plural": "osa-alueet"
+            },
+        ],
+        "closed": True,
+        "organization": None,
+        "geojson": None,
+        "main_image": None,
+        "contact_persons": [{
```
Rikuoja: What happens if we POST a contact person that is not present in the backend? I guess the NestedPKRelatedField does not allow creating new objects and specifying their PK?  This means the other fields are redundant when posted, because this does not allow changing or creating a contact person?

tuomas777: You cannot create or modify a contact person there, just select. The contact person is selected by giving it's ID in the object, ie. `"contact_persons": [{"id": <id>}]` You can also post other fields there to ensure that the data is round-trippable, but those don't affect anything.

Rikuoja: Yeah that's what I thought. Probably will have to implement separately POSTing labels and contact persons then.

JsseL: @Rikuoja Yeah, I think we had the idea to begin with "select only" and iterate from that.

PR #247, opened 2016-12-19
Travis: Fix path to requirements, given the demise of requirements-dev.txt

vikoivun:


vikoivun:
Force pushed a version with public key for correct repository.

PR #248, opened 2016-12-21
Fix comment PUT / PATCH / DELETE

### LINKED ISSUE #244: Allow logged-in users to (soft-)DELETE their own comments

tuomas777:
- Made some changes to root comment endpoint viewset which make
  modifying and deleting comments work better.

- Changed hard delete to soft delete.

- Added comment PUT, PATCH and DELETE tests for both the nested and
  the root endpoint.

Closes #244

### DIFF
--- a/democracy/tests/test_comment.py
+++ b/democracy/tests/test_comment.py
@@ -16,6 +16,25 @@ from democracy.tests.utils import (
 )
 
 
+root_list_url = '/v1/comment/'
+
+
+def get_root_detail_url(comment):
+    return root_list_url + str(comment.pk) + '/'
+
+
+def get_nested_detail_url(comment):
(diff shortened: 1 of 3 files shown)


democracy/tests/test_comment.py
```diff
@@ -685,3 +704,83 @@ def test_only_unauthenticated_can_set_author_name(client, can_set_author_name, r
...
+    comment = default_hearing.get_main_section().comments.all()[0]
+    url = get_detail_url(comment)
+
+    response = john_doe_api_client.delete(url)
+    assert response.status_code == 204
+
+    comment = SectionComment.objects.everything().get(id=comment.id)
+    assert comment.deleted is True
+
+
```
Rikuoja: Here, we should add a test that anonymous comments cannot be deleted by any user, and any comments cannot be deleted by an anonymous user.

PR #250, opened 2016-12-29
Add language support

stephawe:
Add django parler
Make model fields translatable
Add migration for django-parler
Update the tests to take into account the translation
Add a Translatable serializer
Prevent image duplications
Fix the copy_hearing function to deal with translation
Update the admin to work with translations
Add a serializer for CommentImage
Fix hearing reports to work with translations
Add language detection to comments
Update serializers to use the TranslatableSerializer
Serialize the comment language code
Translate the Hearing abstract


democracy/migrations/0030_migrate_translatable_fields.py
```diff
@@ -0,0 +1,78 @@
...
+        MyModel = apps.get_model('democracy', modelname)
+        MyModelTranslation = apps.get_model('democracy', '%sTranslation' % modelname)
+        model_fields = [f.name for f in MyModelTranslation._meta.get_fields()]
+
+        for object in MyModel.objects.all():
+            try:
+                translation = _get_translation(object, MyModelTranslation)
+                assign(object, translation, model_fields)
+                object.save()   # Note this only calls Model.save() in South.
+            except:
```
tuomas777: This hides all possible errors, wouldn't it be better to catch just expected exceptions?

democracy/views/utils.py
```diff
@@ -199,3 +203,112 @@ def to_internal_value(self, data):
...
+    """
+    A serializer for translated fields.
+
+    translated_fields must be declared in the Meta class.
+    By default, translation languages obtained from settings, but can be overriden
+    by defining translation_lang in the Meta class.
+    """
+
+    def __init__(self, *args, **kwargs):
+        assert hasattr(self.Meta, 'translated_fields'), 'Improperly configured: Need to specify translated_fields'
```
Rikuoja: Why is there a need for listing the field separately in `fields` and `translated_fields` in Meta? Surely, the model contains information on which fields are translated, so we can add to `translated_fields` all those fields that are listed in `fields` and are contained in translations?

democracy/models/comment.py
```diff
@@ -61,6 +61,8 @@ def save(self, *args, **kwargs):
             raise ValueError("Comments must have either plugin data, textual content or label")
         if not self.author_name and self.created_by_id:
             self.author_name = (self.created_by.get_display_name() or None)
+        if not self.language_code and self.content:
+            self.language_code = detect(self.content)
```
stephawe: Good point. I have updated the PR accordingly.

democracy/views/utils.py
stephawe: `translated_fields` have the format `{field: {"en": "something", "fi": "..."}}`. If the field is not in the `translated_fields` (despite having a translation), the serialization would be `{field: "something"}`.

This might not very useful though. I can remove it if it is useless and makes things too complicated.

democracy/models/comment.py
```diff
@@ -56,11 +57,25 @@ def parent_id(self):
         """
         return getattr(self, "%s_id" % self.parent_field, None)
 
+    def _detect_lang(self):
+        try:
+            candidates = detect_langs(self.content)
+            for candidate in candidates:
+                if candidate.lang in [trans.language_code for trans in self.section.translations.all()]:
```
Rikuoja: Actually, let's change this to `settings.PARLER_LANGUAGES`, because we should actually allow (and already do have) comments in all supported languages even though the hearing only is in Finnish.

democracy/models/comment.py
```diff
@@ -56,11 +57,25 @@ def parent_id(self):
         """
         return getattr(self, "%s_id" % self.parent_field, None)
 
+    def _detect_lang(self):
+        try:
+            candidates = detect_langs(self.content)
```
Rikuoja: Noticed that the stupid language detection library is not case-agnostic. Uppercase Finnish get detected as Vietnamese of all things :D Therefore, let's `self.content.lower()` for all content.

democracy/migrations/0032_add_language_code_to_comment.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AlterModelOptions(
+            name='sectionimage',
+            options={'ordering': ('ordering', 'translations__title'), 'verbose_name': 'section image', 'verbose_name_plural': 'section images'},
+        ),
+        migrations.AddField(
+            model_name='sectioncomment',
+            name='language_code',
+            field=models.CharField(blank=True, max_length=15, verbose_name='language code'),
+        ),
```
Rikuoja: Let's add language detection of existing comments (mostly in Finnish, Swedish and English) here, so just run `comment._detect_lang()` for all the data?

democracy/migrations/0030_migrate_translatable_fields.py
```diff
@@ -0,0 +1,78 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.9.10 on 2016-12-08 12:00
+from __future__ import unicode_literals
+
+from django.conf import settings
+from django.db import migrations
+from django.core.exceptions import ObjectDoesNotExist
+
+IMPORT_LANGUAGE_CODE = 'fi'  # Assume that the existing content is in Finnish
```
Rikuoja: Actually there are one or two Hearings in Swedish or English already present. Could we pick the Section, Hearing and SectionImage language using `detect_lang`? Labels and ContactPerson don't have long enough strings for the detection to be reliable.

stephawe: It is possible, but if there is only a few non Finnish Hearings, I would recommend against since the detection may be incorrect or even fail, leading us with probably more Hearing to fix manually.

democracy/models/comment.py
stephawe: Nice :)
I have forced the detection to lowercase content as you suggested.

democracy/migrations/0032_add_language_code_to_comment.py
stephawe: Added

PR #251, opened 2017-01-09
Allow user to have a nickname

### LINKED ISSUE #207: Allow registered user to save nickname

stephawe:
Add a field nickname to user 
Test that the nickname is used when a user create a comment

closes #207


Rikuoja:
To implement desired UI functionality, we actually need the user to be able to specify their nickname when POSTing a comment.

Hence, allow posting `author_name` when a registered user POSTs a comment, much as when an unregistered user POSTs a comment.

Then, if a registered user POSTs a comment with `author_name`, actually change the user `nickname` accordingly, replacing the previous one. This way, the next time the user tries to POST a comment in the UI, they will see it as their default nickname.

democracy/views/section_comment.py
```diff
@@ -42,6 +42,10 @@ def to_internal_value(self, data):
             data["plugin_data"] = ""
         if data.get("images") is None:
             data["images"] = []
+        user = self.context['request'].user
+        if user and data.get('author_name'):
+            user.nickname = data['author_name']
+            user.save()
```
tuomas777: Also, IMO `to_internal_value()` seems a bit wrong place for this, maybe `save()` ?

Rikuoja:
LGTM!

PR #252, opened 2017-01-13
Update the HearingFilter and LabelFilter for translations

stephawe:


Rikuoja:
Wow, that was simpler than I thought :)

PR #263, opened 2017-09-19
Try to fix failing travis test

Rikuoja:


Rikuoja:
* Discarding the problematic django-parler cache for the time being.
* Optimized prefetches at the hearing endpoint (which is the one the UI uses the most) results in faster API response even without the cache.

democracy/views/hearing.py
```diff
@@ -253,8 +253,9 @@ class HearingListSerializer(HearingSerializer):
 
     def get_fields(self):
         fields = super(HearingListSerializer, self).get_fields()
-        # Elide section and geo data when listing hearings; one can get to them via detail routes
+        # Elide section, contact person and geo data when listing hearings; one can get to them via detail routes
         fields.pop("sections")
+        fields.pop("contact_persons")
```
tituomin: Breaking change: I assume there are no API users who might use this field.

Rikuoja: Nope, the API is actually not published yet so the only user is the UI itself. Contact persons are not rendered in the event listing, so it actually makes no sense to fetch them from the database at this point.

PR #274, opened 2017-11-15
Pull configuration from environment

vikoivun:
This uses django-environment to read from environment.
Also pulls raven in unconditionally. Raven configuration still to be done.

### DIFF
--- a/kerrokantasi/settings/base.py
+++ b/kerrokantasi/settings/base.py
@@ -10,14 +10,19 @@ BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__fil
 env = environ.Env()
 
 DEBUG = env.bool('DEBUG', default=True)
-SECRET_KEY = env.str('SECRET_KEY', default=('x' if DEBUG else environ.Env.NOTSET))
+SECRET_KEY = env.str('SECRET_KEY', default=('x' if DEBUG else ''))
 ALLOWED_HOSTS = env.list('ALLOWED_HOSTS', default=[])
-ADMINS = env.list(default=[])
+ADMINS = env.list('ADMINS', default=[])
 DATABASES = {
     'default': env.db('DATABASE_URL', default='postgis:///kerrokantasi')
 }
 SENTRY_DSN = env.str('SENTRY_DSN', default='')
 DEMOCRACY_UI_BASE_URL = env.str('DEMOCRACY_UI_BASE_URL', default='http://localhost:8086')
+
+JWT_AUTH = {
+    'JWT_SECRET_KEY': env.str('JWT_SECRET_KEY', default='kerrokantasi'),
+    'JWT_AUDIENCE': env.str('JWT_AUDIENCE', default='kerrokantasi')
+}
 
 ### Settings below do not usually need changing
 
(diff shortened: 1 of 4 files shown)


kerrokantasi/settings/base.py
```diff
@@ -1,12 +1,30 @@
...
 gettext = lambda s: s
 
 BASE_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 
-SECRET_KEY = '52k^*)c*bz9t0lzsf_$a+jl3zcy6re!gnw77__)y(#v91-p%tp'
-DEBUG = True
+env = environ.Env()
 
-ALLOWED_HOSTS = []
+DEBUG = env.bool('DEBUG', default=True)
```
akx: Is `True` a sane default for `DEBUG`? It sounds like that makes it possible for it to be inadvertently on in production.

kerrokantasi/settings/base.py
```diff
@@ -30,6 +48,7 @@
     'democracy',  # Reusable participatory democracy app
     'parler',
     'django_filters',
+    ''
```
akx: What's this?

kerrokantasi/settings/base.py
```diff
@@ -103,15 +114,8 @@
     'TEST_REQUEST_DEFAULT_FORMAT': 'json',
 }
 
+JWT_AUTH['JWT_PAYLOAD_GET_USER_ID_HANDLER'] = 'helusers.jwt.get_user_id_from_payload_handler'
```
akx: Why is this separately here? It could be initialized in the `JWT_AUTH` dict with the rest...

kerrokantasi/settings/base.py
```diff
@@ -1,12 +1,30 @@
...
+}
+SENTRY_DSN = env.str('SENTRY_DSN', default='')
+DEMOCRACY_UI_BASE_URL = env.str('DEMOCRACY_UI_BASE_URL', default='http://localhost:8086')
+
+JWT_AUTH = {
+    'JWT_SECRET_KEY': env.str('JWT_SECRET_KEY', default='kerrokantasi'),
+    'JWT_AUDIENCE': env.str('JWT_AUDIENCE', default='kerrokantasi')
+}
+
+### Settings below do not usually need changing
```
akx: This is something of a moot comment since you're probably not supposed to modify this file locally anyway, just use envvars.

kerrokantasi/settings/base.py
```diff
@@ -1,12 +1,30 @@
 import os
+
+import environ
+import raven
```
akx: Soooo, why exactly _is_ this imported?

kerrokantasi/settings/base.py
vikoivun: Still this file serves as a template, when using local_settings (which is still wanted)

kerrokantasi/settings/base.py
vikoivun: Missing (WIP) tag from the PR. Ie. this is still WIP.

kerrokantasi/settings/base.py
Rikuoja: The developer might override these in `local_settings.py`, though.

kerrokantasi/settings/base.py
vikoivun: This is not really a configurable thing. Ie. configurables are at top and non-configurables sit beyond the veil.

kerrokantasi/settings/base.py
vikoivun: Good point. I just translated this.

PR #277, opened 2017-11-20
Issue 265

### LINKED ISSUE #265: Fix django-admin for contact persons
Currently, contact persons cannot be edited in the admin due to server error (HTTP 500).

sanesane:
fix #265

### DIFF
--- a/democracy/admin/__init__.py
+++ b/democracy/admin/__init__.py
@@ -225,7 +225,7 @@ class LabelAdmin(TranslatableAdmin, admin.ModelAdmin):
     exclude = ("published",)
 
 
-class SectionTypeAdmin(TranslatableAdmin, admin.ModelAdmin):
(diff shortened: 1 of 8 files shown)


sanesane:
This was not admin specific problem, the root cause was how model inheritance hierarchy was misused regarding TranslatableModels, causing The Diamond Problem. Specifically, the default model manager "objects" was shadowed in a way that translatable model forms broke in admin panel. However, the root cause affected all application code using these models.

Testing admin pages will not prevent this problem to happen again.

Added tests for Organization admin changeist and change pages, which was the only model in democrazy app which was missing tests in previous commits. It doesn't use TranslatableModel though.

That being said, it doesn't hurt to have simple "renders ok" test for all admin views, even those that aren't in democracy app, but could this be separate issue.

PR #280, opened 2017-11-27
Issue 276

### LINKED ISSUE #276: When POSTing or PUTting a Hearing, require a non-empty title (in at least one language) and a non-empty slug
Currently, the user may unintendedly POST or PUT a Hearing with an empty title or slug string.

Apparently a unique slug or slug based on a title is not automatically generated (and probably couldn't be generated if the title is empty, too). This results in the created Hearing being inaccessible in the frontend. Fix tests to require non-empty title and slug at POST and PUT.

sanesane:
fix #276

### DIFF
--- a/democracy/views/hearing.py
+++ b/democracy/views/hearing.py
@@ -69,18 +69,16 @@ class HearingCreateUpdateSerializer(serializers.ModelSerializer, TranslatableSer
         self.partial = kwargs.get('partial', False)
 
     def validate(self, data):
-        # Seems like TranslatableFields always allows null values by design, so title can't
-        # be set as null=False in model.
-        # DRF does not use model validation routine at all,
-        # so this check can't be in model.clean()-method.
-        # For some reason per field validation in this serializer is not working with parler
+        data = super(HearingCreateUpdateSerializer, self).validate(data)
+        action = self.context['view'].action
         try:
-            # When creating or update as whole, there must be title
             titles = data['title']
         except KeyError:
-            raise serializers.ValidationError('Title is required at least in one locale')
+            # When creating or updating as whole, there must be title
+            if action in ('create', 'update'):
+                raise serializers.ValidationError('Title is required at least in one locale')
         else:
-            # Title must be set atleast in one locale
+            # When creating or updating as whole or patching, title must be set at least in one locale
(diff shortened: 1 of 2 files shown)


democracy/views/hearing.py
```diff
@@ -68,6 +69,34 @@ def __init__(self, *args, **kwargs):
         super(HearingCreateUpdateSerializer, self).__init__(*args, **kwargs)
         self.partial = kwargs.get('partial', False)
 
+    def validate(self, data):
+        data = super(HearingCreateUpdateSerializer, self).validate(data)
+        action = self.context['view'].action
+        if self._validate_title_present(action, data):
+            return data
+        raise serializers.ValidationError()
```
Rikuoja: Is this row ever reached? I guess the above `if` always returns `True` or raises `ValidationError`.

democracy/views/hearing.py
```diff
@@ -68,6 +69,34 @@ def __init__(self, *args, **kwargs):
...
+            if action in ('create', 'update'):
+                raise serializers.ValidationError('Title is required at least in one locale')
+        else:
+            # When creating or updating as whole or patching, title must be set at least in one locale
+            has_title = False
+            for title in titles.values():
+                if title:
+                    has_title = True
+                    break
+            if not has_title:
```
Rikuoja: Here, we can have just 
```
for title in titles.values():
    if title:
        break
else:
```

so there should be no need for `has_title`.

democracy/views/hearing.py
```diff
@@ -68,6 +69,34 @@ def __init__(self, *args, **kwargs):
...
+            return data
+        raise serializers.ValidationError()
+
+    def _validate_title_present(self, action, data):
+        # Kind of hackish slution to require title when POSTing or PUTing
+        # Parler makes it hard to use build in validations for this, as translatable fields
+        # are always nullable bu default
+        try:
+            titles = data['title']
+        except KeyError:
```
akx: How about just

```python
titles = data.get('title')
if action in ('create', 'update') and not titles:
  raise ...  # no dict or empty dict
if not any(title for title in titles.values()):
  raise ...  # no value was truthy in dict
return True
```

PR #281, opened 2017-11-28
Respect SectionImage ordering field

### LINKED ISSUE #273: Ascertain that `v1/section` images array returns Images in the order specified by the backend `ordering` field
Currently, the `images` array of the `v1/section` json may not be ordered. This may result in the UI showing the wrong image first in the service, if several images are added to the same section in the future.

indigane:
Resolves #273

### DIFF
--- a/democracy/models/section.py
+++ b/democracy/models/section.py
@@ -89,9 +89,7 @@ class Section(Commentable, StringIdBaseModel, TranslatableModel):
 
 
 class SectionImageManager(TranslatableManager, BaseModelManager):
-
-    def get_queryset(self):
-        return super(SectionImageManager, self).get_queryset().order_by('pk')
+    pass
 
 
 class SectionImage(BaseImage, TranslatableModel):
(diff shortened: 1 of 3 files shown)


democracy/views/utils.py
```diff
@@ -105,9 +105,6 @@ def to_representation(self, images):
         else:
             images = images.public()
 
-        # Remove duplicated rows
```
Rikuoja: Was this addition ever necessary? It seems this has been added *after* the `ordering` field was defined for the `SectionImage` model, so surely it still does something? Apparently the reason was "Getting images throug the related name leads to return several time the same images, one per translation (due to the join). This commit orders the images by pk which remove the duplication."

So should we still have `.order_by('ordering')` or just `.distinct()` here?

democracy/models/section.py
```diff
@@ -89,9 +89,7 @@ def plugin_implementation(self):
 
 
 class SectionImageManager(TranslatableManager, BaseModelManager):
-
-    def get_queryset(self):
-        return super(SectionImageManager, self).get_queryset().order_by('pk')
+    pass
```
Rikuoja: As commented below, should we still have `.distinct()` or something in the `get_queryset` here, if this is referred to in a section by the related name and the titles are repeated in several language? Or is it not needed?

PR #283, opened 2017-12-12
Issue 272

### LINKED ISSUE #272: Allow image uploads linked to a section
Similar to https://github.com/City-of-Helsinki/kerrokantasi/issues/266 , allow POST and PUT at the root level Image endpoint. Currently, Images can only be added when inlined in a Section.

This means the user can also add Images to a Section one by one. A SectionImage newly added to an existing Section should have the ordering set to the ordering of an existing SectionImage + 1.

sanesane:
fix #272

### DIFF
--- a/democracy/views/section.py
+++ b/democracy/views/section.py
@@ -18,7 +18,7 @@ from democracy.views.utils import (
 class SectionImageSerializer(BaseImageSerializer, TranslatableSerializer):
     class Meta:
         model = SectionImage
-        fields = ['id', 'title', 'url', 'width', 'height', 'caption', 'ordering']
+        fields = ['id', 'title', 'url', 'width', 'height', 'caption',]
 
 
 class SectionImageCreateUpdateSerializer(BaseImageSerializer, TranslatableSerializer):
--- a/democracy/views/section.py
+++ b/democracy/views/section.py
@@ -1,5 +1,5 @@
 import django_filters
-from django.db.models import Q
+from django.db.models import Q, Max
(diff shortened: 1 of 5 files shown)


Rikuoja:
It's a bit funny IMO that `ordering` is returned explicitly at the root level `image` endpoint, but not when images are inlined in sections and hearings. Makes sense, though, because it is not needed when the json can just be ordered accordingly; but I guess it is better we return the integer explicitly in the global list of all images?

democracy/views/section.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, *args, **kwargs):
 
     class Meta:
         model = SectionImage
-        fields = ['title', 'url', 'width', 'height', 'caption', 'image']
+        fields = ['title', 'url', 'width', 'height', 'caption', 'image', 'ordering']
```
sanesane: Image order wasn't saved to db because ordering field was missing from serializer, so seems like all section images in production have default order of one (1). This should be checked and fixed if so.

Rikuoja: I believe the numbers in production are alright, because multiple images haven't really been POSTed yet through the API.

However, I think this field should only be present at the root level image endpoint. That is because in the section json, the image array should always be ordered according to this field. Similarly, the field should be set when POSTing inlined images according to image order in the POSTed array.

Granted, it's stupid to have the field only in some views, but in most cases it is indeed redundant, as we are only dealing with lists of images within one section.

PR #314, opened 2018-04-05
Security upgrade for Django (1.11.2)

vikoivun:


vikoivun:
Whoops. Already pushed to master.

PR #315, opened 2018-04-23
security hotfix: update Django to 1.11.12

vikoivun:


vikoivun:
Merging forcefully, as there one stochastic test decided to fail

PR #327, opened 2019-01-23
Feature/dockerization

halarotu:
New Dockerfile and an example of local_settings.py


vikoivun (changes requested):
You should leave out local_settings.py.example completely

Instead you should refactor kerrokantasi/settings/base.py to read /config_dev.toml (ie. read from root) for settings. (env.read_env)

Then you just need to write a config_dev.toml.example which explains all the settings and gives working examples. ;)

PR #330, opened 2019-05-10
Refactor config and add support for configuring for protected files

vikoivun:
this needs some verication still. It is now in the nice linkedevents style, where it tells out if if read config from file.


vikoivun:
I think this is camera ready now.

PR #333, opened 2019-08-30
Make map settings configurable

tomijarvi:
The default coordinate will be Helsinki's coordinate as it has been used past.


tomijarvi:
@Rikuoja: There is some weird issue in the docker build. Do you know anything about it? If that is something else, this PR can be a review so we can get this map coordinates configuration to Turku's Kerrokantasi.

Rikuoja:
Hm. This looks something I have seen before, probably @jussiarpalahti or @vikoivun know why the travis build stalls when docker asks for password?

PR #334, opened 2019-09-03
Don't containerize when docker password missing

jussiarpalahti:


jussiarpalahti:
Since PR check didn't get the password and push check does (since I'm a contributor) I'd say this works, @Rikuoja

PR #338, opened 2019-11-06
Configure authentication to use OpenId Connect

Jogha:
Remove old token authentication related configurations from settings.


Rikuoja:
Two or three things before merging this:

1) To enable silent renew, due to the current oidc library limitations, the OIDC_LEEWAY value will have to be set the same as the token age, like e.g. in https://github.com/City-of-Helsinki/kukkuu/commit/9fc249e14a2fc5b72a970bb8090ca31f51f44d1c

2) Add some explanations to config_dev.toml.example so that Tunnistamo user can easily deduce which strings to fill in which setting, now that there are many of them. Probably @vikoivun can also suggest suitable help texts for the settings.

3) One way or the other, fix the admin template load problem that happens if `django.contrib.admin` is in INSTALLED_APPS after the new apps.

Other than that, however, this seems to work splendidly, got it to work with the UI branch, so would be happy to merge this!

PR #373, opened 2022-01-17
Fix deleted comment messages

ranta:


democracy/views/section_comment.py
```diff
@@ -135,24 +135,26 @@ class SectionCommentSerializer(BaseCommentSerializer):
...
-        if obj.deleted_by_id == obj.created_by_id:
+        elif obj.deleted_by_id is not None and obj.deleted_by_id == obj.created_by_id:
             return "Kirjoittaja poisti oman viestinsä."
-
-        return f"Viesti on poistettu {obj.deleted_at.strftime('%-d.%-m.%Y %H:%M')}, " \
-               f"koska se ei noudattanut Kerrokantasi-palvelun sääntöjä https://kerrokantasi.hel.fi/info"
+        elif obj.deleted_at:
+            deleted_time = f" {obj.deleted_at.strftime('%-d.%-m.%Y %H:%M')}" if obj.deleted_at is not None else ""
+            return f"Viesti on poistettu{deleted_time}, " \
+                   f"koska se ei noudattanut Kerrokantasi-palvelun sääntöjä https://kerrokantasi.hel.fi/info"
```
vikoivun, line 156: Would it still be possible to change this to use DEMOCRACY_UI_BASE_URL setting to refer to the UI?

vikoivun (approved):

democracy/views/section_comment.py
ranta: Good catch, I didn't realise this setting even existed.
We need to remember to also patch the hardcoded URL from the frontend translations

vikoivun: Nice.

PR #381, opened 2022-07-05
Upgrade to Django 3, Automated code formatting

ranta:
tl;dr:
Bump Django to 3 and dependencies to versions that work with it, apply automated formatting on many files.

- Bump Django version from outdated 2.2 LTS to 3.2 LTS, which has support until April 2024.
- Also bump dependencies to versions supported by Django 3
- And also bump dev-dependency versions
- Fix deprecation issues caused by the upgrade 👴 
- Add linter config for `black`
- Run `isort` on all files. This will make imports look beautiful 🌹 and hopefully reduce conflicts in the future
- Run file formatting using `black . --skip-string-normalization`
  - Skip file formatting, as this project has mostly single quotes, and I wanted to keep the diff to a minimum. Black could be run later without the option to skip string normalization though 🤔 
  - This will also help keep future conflicts lower and of course, make the code easier to read 🤓 

refs. KER-74

Adding GHA CI to check for code style & running tests might be a good next step, but it will be left for a later date.


vikoivun (approved):
Look like sensible housekeeping to me

PR #386, opened 2022-08-15
Preserve contact person order in hearings

ranta:
Preserve the order for contact persons, in which they are sent to the API in.
This allows for listing more important people to the hearing first, instead of results being in alphabetical order.
refs. KER-50


SanttuA (approved):

ranta:
Due to implicit joins the contact persons were being duplicated in e.g. Django admin. Force pushed a fix to it

PR #401, opened 2023-03-22
Fix running custom commands on the container

charn:
Remove the "e" option for running custom commands on the container. Instead we use the expected default behavior/convention that if the container is provided with an unrecognized command it will try to run it.

This change will allow OpenShift terminal to run and also allow PyCharm to run its own commands.

Also instead of checking for exitcode explicitly we use `set -e` in scripts which will cause the entrypoint script to automatically exit if it encounters a non-zero exit code.

As a final this `shellcheck` was run on entrypoint related scripts.


voneiden (approved):
deploy/run_maintenance.sh is a bit dead weight at the moment at least but LGTM

I'm used to just defining kubernetes cronjobs that call management commands directly.

Jurppe (approved):
Much cleaner this way, nice 👍

PR #402, opened 2023-03-30
Soft delete comment images when comment is soft deleted

voneiden:
Previously images were not soft deleted when a comment was soft deleted. This PR adds soft deletion to also images and improves test coverage to check that both poll answers and images related to a comment are soft deleted.

Additionally removed unused usage of `using` from various places as it seemed unused.

https://helsinkisolutionoffice.atlassian.net/browse/KER-149


danipran (approved):
LGTM :+1:

PR #405, opened 2023-05-02
KER-167: Add can_delete serializer field in SectionCommentSerializer

danipran:
Add `can_delete` serializer method field in `SectionCommentSerializer` which tells whether the comment can be deleted or not.

Currently, the `can_edit` field covers both editing *and* deleting, so a more correct name would've been `can_delete_and_edit`. This PR separates these to their own fields, since they have different conditions. This is required for front-end representation; now it's all or nothing, but we want something from in-between as well.

Also move methods to their proper places, i.e. don't handle `SectionComment` stuff in `BaseComment`.

Related front-end PR: https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/932

### DIFF
--- a/democracy/views/section_comment.py
+++ b/democracy/views/section_comment.py
@@ -2,6 +2,7 @@ import django_filters
 from django.conf import settings
 from django.core.exceptions import ValidationError as DjangoValidationError
 from django.db.transaction import atomic
+from django.utils.functional import cached_property
 from django.utils.translation import ugettext as _
 from rest_framework import filters, response, serializers, status
 from rest_framework.exceptions import ValidationError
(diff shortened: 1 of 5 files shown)


democracy/views/section_comment.py
```diff
@@ -231,6 +236,16 @@ def get_deleted_by_type(self, obj):
         # No information about who deleted the comment
         return "unknown"
 
+    @cached_property
+    def request(self):
+        return self.context.get('request', None)
```
tuomas777, line 241: Unnecessary default value.
```suggestion
        return self.context.get('request')
```

tuomas777, line 241: Unnecessary default value.
```suggestion
        return self.context.get('request')
```

tuomas777 (approved):
Really nice that this also makes the permission stuff more DRY, good job! 👍 🚀

PR #406, opened 2023-05-09
check if the key exists in request.META object before accessing it

Jurppe:
Simple request such as
```curl
GET /v1/file/
```

would cause key error since it does not provide `CONTENT_TYPE` in metadata.


danipran (approved):
LGTM :+1:

PR #409, opened 2023-06-08
Reduce uwsgi worker count

Jurppe:
Reduce default uwsgi worker processes count. Scaling strategy scales pods in cluster horizontally.


danipran (approved):
LGTM :+1:

PR #411, opened 2023-07-25
KER-203 | Hide author column from excel-file

tonipel:
Hide author column from excel-file for staff users.

Refs KER-203


danipran (approved):
LGTM :+1:

PR #424, opened 2023-08-14
Ker-193

arpargo:
Deleted comments were not shown in frontend if they were only reply to another comment or if all other replies were also deleted.

### DIFF
--- a/democracy/tests/conftest.py
+++ b/democracy/tests/conftest.py
@@ -195,6 +195,35 @@ def hearing_without_comments(contact_person, default_organization, default_proje
 
     return hearing
 
+@pytest.fixture()
+def hearing_with_comments_on_comments(jane_doe, john_doe, contact_person, default_organization, default_project):
+    """
+    Fixture for a simple hearing with comments that have replies.
+    """
+    hearing = Hearing.objects.create(
+        title='Simple hearing without comments',
+        open_at=now() - datetime.timedelta(days=1),
+        close_at=now() + datetime.timedelta(days=1),
+        slug='simple-hearing-slug',
+        organization=default_organization,
+        project_phase=default_project.phases.all()[0],
+    )
+
+    section_type = InitialSectionType.MAIN
+    section = Section.objects.create(
+        abstract='Section abstract for simple hearing',
+        hearing=hearing,
+        type=SectionType.objects.get(identifier=section_type),
+        commenting=Commenting.OPEN,
+    )
+
+    assert_ascending_sequence([s.ordering for s in hearing.sections.all()])
+    hearing.contact_persons.add(contact_person)
+    parent_comment = section.comments.create(created_by=john_doe, content=default_comment_content[::-1])
+    parent_comment.comments.create(created_by=jane_doe, content=default_comment_content[::-1], section=section)
+    parent_comment.comments.create(created_by=john_doe, content=default_comment_content[::-1], section=section)
+    return hearing
+
 
 @pytest.fixture()
 def strong_auth_hearing(john_doe, contact_person, default_organization, default_project):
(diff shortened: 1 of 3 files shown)


voneiden:
I generated ~300 comments, some of them replies, some not and it looks like that with page_size 50 a simple GET of the comments endpoint results in

`70.55 ms (212 queries including 206 similar and 56 duplicates ) ` 

Unexpectedly it's actually better than  `master`

` default 91.35 ms (214 queries including 208 similar and 58 duplicates ) `

But we would now have a great opportunity to improve this.

voneiden:
The duplicate queries for each comment are (there are 4 of them)

1) `SELECT ••• FROM "democracy_commentimage" WHERE (NOT ("democracy_commentimage"."deleted") AND "democracy_commentimage"."comment_id" = 289) ORDER BY "democracy_commentimage"."ordering" ASC, "democracy_commentimage"."title" ASC    `

2) `SELECT ••• FROM "democracy_section" WHERE "democracy_section"."id" = '4bCl2TMEHoxiiTtdGIH7QEXzGhv4OwG2' LIMIT 21`

3) `SELECT ••• FROM "democracy_sectionpollanswer" INNER JOIN "democracy_sectionpolloption" ON ("democracy_sectionpollanswer"."option_id" = "democracy_sectionpolloption"."id") INNER JOIN "democracy_sectionpoll" ON ("democracy_sectionpolloption"."poll_id" = "democracy_sectionpoll"."id") WHERE (NOT ("democracy_sectionpollanswer"."deleted") AND "democracy_sectionpollanswer"."comment_id" = 274)`

4) `SELECT ••• FROM "democracy_sectioncomment" WHERE "democracy_sectioncomment"."comment_id" = 287 ORDER BY "democracy_sectioncomment"."created_at" DESC`

voneiden:
We can also consider separating perf improvements into a separate PR, but this kind of approach works for the comments... it's does get a wee bit mind boggling though:

in `get_queryset` we do

```python
queryset = queryset.prefetch_related(Prefetch("comments", SectionComment.objects.everything().only("pk")))
```

```python
    def get_comments(self, obj):
        queryset = [c.pk for c in obj.comments.all()]
        return queryset
```

this yields a single SQL that looks like


```sql
SELECT "democracy_sectioncomment"."id"   FROM "democracy_sectioncomment"  WHERE  "democracy_sectioncomment"."comment_id" IN (317, 316, 315, 314, 313,  312, 311, 310, 309, 308, 307, 306, 305, 304, 303, 302, 301, 300, 299,  298, 297, 296, 295, 294, 293, 292, 291, 290, 289, 288, 287, 286, 285,  284, 283, 282, 281, 280, 279, 278, 277, 276, 275, 274, 273, 272, 271,  270, 269, 268)  ORDER BY "democracy_sectioncomment"."created_at" DESC
```

Other get_queryset improvements are

```python 
queryset = queryset.select_related("section").prefetch_related("images")
queryset = queryset.prefetch_related(Prefetch("poll_answers", SectionPollAnswer.objects.select_related("option__poll")))
```

this brings the queries down to something quite a bit more reasonable:

14.58 ms (15 queries including 6 similar and 6 duplicates )

arpargo:
I hope I understood the suggestions properly and made the correct optimizations. @voneiden

voneiden:
Sorry, my notes were ambiguous. The changes need to be made on `CommentViewSet` in `section_comment.py` instead of `BaseCommentViewSet` in `comment.py`. While the former inherits the latter, it also overrides the `get_queryset` and does not call super(), so the changes introduced in 2ab9431 `BaseCommentViewSet` have no effect - at least in comment listing. 

I'm totally fine if we just split the perf issues from this PR into a separate issue, and then you can just drop the 2ab9431 and get this merged.

voneiden:
split https://helsinkisolutionoffice.atlassian.net/browse/KER-246

voneiden (approved):

danipran (approved):
LGTM :+1:

PR #426, opened 2023-08-15
Bump easy-thumbnails

charn:
New version supports the new pillow library.


voneiden (approved):
:chart_with_upwards_trend:

nicobav (approved):
Alrighty 👍 

I didn't notice any usage of settings: `THUMBNAIL_HIGH_RESOLUTION` and `THUMBNAIL_HIGHRES_INFIX ` which are dropped as from 2.8.0. Just a noting those here aloud if those rings any bells.

PR #433, opened 2023-09-07
KER-202 KER-239 | Contact person endpoint can only be accessed by admins

charn:
This PR add more strict permissions for the contact person endpoint. Only authenticated organization admins will be able to use the entpoint. Also permission checks and related responses are refactored (for the viewset in question) to be more standard Django restframework.

### DIFF
--- a/democracy/views/contact_person.py
+++ b/democracy/views/contact_person.py
@@ -1,8 +1,15 @@
-from rest_framework import mixins, permissions, response, serializers, status, viewsets
+from rest_framework import mixins, permissions, serializers, viewsets
 
 from democracy.models import ContactPerson, Organization
 from democracy.pagination import DefaultLimitPagination
 from democracy.views.utils import TranslatableSerializer
+
+
+class ContactPersonPermission(permissions.BasePermission):
+    message = "User without organization cannot access contact persons."
+
+    def has_permission(self, request, view):
+        return bool(hasattr(request.user, "get_default_organization") and request.user.get_default_organization())
 
 
 class ContactPersonSerializer(serializers.ModelSerializer, TranslatableSerializer):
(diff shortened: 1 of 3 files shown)


democracy/views/contact_person.py
```diff
@@ -1,10 +1,18 @@
...
 from democracy.models import ContactPerson, Organization
 from democracy.pagination import DefaultLimitPagination
 from democracy.views.utils import TranslatableSerializer
 
 
+class ContactPersonPermission(IsAuthenticated):
+    message = "User without organization cannot access contact persons."
+
+    def has_permission(self, request, view):
+        return bool(super().has_permission(request, view) and request.user.get_default_organization())
```
danipran, line 13: If you make custom permissions, inherit them from `BasePermission`. Inheriting from anything else can obfuscate the code and can act as a potential source of spaghetti code.

I assume the authenticatio part doesn't really matter for this permission either; what matters is that the method exists. So could do it something like this instead:
```suggestion
class ContactPersonPermission(permissions.BasePermission):
    message = "User without organization cannot access contact persons."

    def has_permission(self, request, view):
        return bool(hasattr(request.user, "get_default_organization") and request.user.get_default_organization())
```

democracy/views/contact_person.py
```diff
@@ -15,26 +23,12 @@ class Meta:
...
-            validated_data["organization"] = self.context['request'].user.get_default_organization()
+            validated_data["organization"] = self.context["request"].user.get_default_organization()
         return super().create(validated_data)
 
 
 class ContactPersonViewSet(viewsets.ReadOnlyModelViewSet, mixins.CreateModelMixin, mixins.UpdateModelMixin):
     serializer_class = ContactPersonSerializer
     queryset = ContactPerson.objects.select_related("organization").order_by("name")
-    permission_classes = (permissions.IsAuthenticatedOrReadOnly,)
+    permission_classes = (ContactPersonPermission,)
```
danipran, line 33: Related to my other comment on `ContactPersonPermission`, you could do this instead:
```suggestion
    permission_classes = [permissions.IsAuthenticated, ContactPersonPermission]
```
Or:
```suggestion
    permission_classes = [permissions.IsAuthenticated & ContactPersonPermission]
```

danipran (changes requested):

danipran (approved):
LGTM :+1:

PR #436, opened 2023-09-12
Optimize SectionComment viewset queries

voneiden:
A bunch of tweaks to reduce the number of db queries significantly.

Also adds a simple test to keep an eye on query numbers


danipran (approved):
LGTM :+1: some of the tests are failing, but not due to these changes; rebasing to `master` seems to fix them.

PR #438, opened 2023-09-12
KER-245 | Disable uwsgi write errors

nicobav:
Disables write errors which occurs when client closes connection

Refs KER-245


danipran (approved):
:+1:

PR #439, opened 2023-09-13
Optimize comment relation translations and browsable API

voneiden:
Need for these came apparent only after using data in the dev environment.

### DIFF
--- a/democracy/views/section_comment.py
+++ b/democracy/views/section_comment.py
@@ -7,12 +7,13 @@ from django.utils.functional import cached_property
 from django.utils.translation import ugettext as _
 from rest_framework import filters, response, serializers, status
 from rest_framework.exceptions import ValidationError
+from rest_framework.relations import PrimaryKeyRelatedField
 from rest_framework.serializers import as_serializer_error
 from rest_framework.settings import api_settings
 from urllib.parse import urljoin
 
 from democracy.enums import Commenting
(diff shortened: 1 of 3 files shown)


democracy/views/section_comment.py
```diff
@@ -161,6 +162,12 @@ class SectionCommentSerializer(BaseCommentSerializer):
     Serializer for comment added to section.
     """
 
+    section = PrimaryKeyRelatedField(
+        queryset=Section.objects.all().prefetch_related(
+            "translations",
+            "hearing__translations",
+        )
+    )
```
voneiden, line 170: This concerns the Browsable API comment POST form, which generates a select for all sections.

democracy/views/section_comment.py
```diff
@@ -437,6 +444,7 @@ class CommentFilterSet(django_filters.rest_framework.FilterSet):
     created_at__lt = django_filters.IsoDateTimeFilter(field_name='created_at', lookup_expr='lt')
     created_at__gt = django_filters.rest_framework.IsoDateTimeFilter(field_name='created_at', lookup_expr='gt')
     comment = django_filters.ModelChoiceFilter(queryset=SectionComment.objects.everything())
+    section = django_filters.CharFilter(field_name='section__id')
```
voneiden, line 447: This concerns the Browsable API filter options. Removes "ModelChoiceFilter" and replaces it with a CharFilter so it's similar to the hearing filter that is on line 442

danipran (approved):
LGTM :+1:

PR #452, opened 2023-10-13
KER-257 | Hide author_name on comment endpoint for unfiltered request

charn:
For privacy reasons the author name is not returned from the API if a filter (hearing, section, comment, created_by) hasn't been used.

For admin users the author name is always returned.

### DIFF
--- a/democracy/tests/integrationtest/test_comment.py
+++ b/democracy/tests/integrationtest/test_comment.py
@@ -1437,6 +1437,16 @@ def test_get_comments_created_by_user(john_doe_api_client, jane_doe_api_client, 
     assert data["results"][0]["content"] == "Jane created this comment"
 
 
+@pytest.mark.parametrize("query_params,expected", [({"created_by": "me"}, 0), (None, 9)])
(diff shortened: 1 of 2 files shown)


nicobav (approved):
Approving since code wise its a ok and does not introduce anything new to api.

The ticket specs allows one be more open minded with the interpretation and I myself started to ponder this. In the specs are mentioned "wide range of audience..." who shouldn't be seeing this field data, however anonymous user can see the field data with inputting query params. I'm wondering should we allow this field only for authenticated users?  As with creator seems to be if i'm reading it correct.

charn:
`author_name` should be available for the UI, for all users. Requiring user to be authenticated is too much as far as I can tell. That's why I haven't hidden the `author_name` in `SectionCommentViewSet`. `SectionCommentViewSet` could be used in the UI in some cases for fetching comments, but that view seems to be missing some filters that exist for `CommentViewSet` which is the root level comment endpoint. I guess one option would be to add those missing filters to that endpoint.

PR #453, opened 2023-11-02
KER-278 | Bump packages alerted by snyk

nicobav:
Snyk reported there to be some sec vurns in django, urllib3, cryptography and pillow in requirements and requirements-dev => let's bump'em.

Refs KER-278


charn (approved):
You could probably also bump those dependabot packages. :+1: :shipit:

PR #455, opened 2023-11-10
feat: add GDPR delete method

voneiden:


charn (approved):
LGTM! :+1: :shipit: I think this can be merged to the draft PR I have open (to get the base for tests). And as discussed you could still check if we should just delete the user or do something regarding the `uuid` field.

PR #456, opened 2023-11-10
KER-273 | Read database password from env

nicobav:
Read db password from env variable if exists

Refs KER-273


charn (approved):
LGTM! :+1: :shipit:

PR #459, opened 2023-12-01
KER-99 | Don't render author when comment deleted

nicobav:
For deleted comments the author information should not be rendered through api

Refs KER-99


danipran (approved):
LGTM :+1:

PR #480, opened 2024-02-16
KER-280 | GDPR additions and changes

nicobav:
Adds following changes to gdpr api get

- Show (soft)deleted comments
- Renders file and image urls instead of bytes
   - Added custom middleware for accessing the request so that the url could be built.

Refs KER-280

### DIFF
--- a/democracy/models/gdpr_data_serialization_mixin.py
+++ b/democracy/models/gdpr_data_serialization_mixin.py
@@ -1,0 +1,14 @@
+from kerrokantasi.utils import get_current_request
+
+
+class FileFieldUrlSerializerMixin:
+    field_to_use_as_url_field = None  # Override in subclass.
+
+    @property
+    def url(self):
+        field = getattr(self, self.field_to_use_as_url_field, None)
+
+        if request := get_current_request():
+            return request.build_absolute_uri(field.url)
+
+        return field.url
--- a/democracy/models/section.py
+++ b/democracy/models/section.py
@@ -22,11 +22,11 @@ from democracy.models.base import (
 )
 from democracy.models.comment import BaseComment, recache_on_save
 from democracy.models.files import BaseFile
+from democracy.models.gdpr_data_serialization_mixin import FileFieldUrlSerializerMixin
 from democracy.models.hearing import Hearing
 from democracy.models.images import BaseImage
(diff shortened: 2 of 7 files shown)


terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr480.api.dev.hel.ninja** :rocket::rocket::rocket:

democracy/models/gdpr_data_serialization_mixin.py
```diff
@@ -0,0 +1,13 @@
+class FileFieldUrlSerializerMixin:
+    field_to_use_as_url_field = None  # Override in subclass.
+
+    @property
+    def url(self):
+        from kerrokantasi.gdpr import _thread_locals
+
+        field = getattr(self, self.field_to_use_as_url_field, None)
+
+        if request := getattr(_thread_locals, "request", None):
```
charn, line 10: I think it would be better if we could hide the implementation details and maybe add a utility function like `get_current_request()` for getting the request.

```
def get_current_request():
    return getattr(_thread_locals, "request", None)
```

nicobav: surely 👍

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr480.api.dev.hel.ninja** :rocket::rocket::rocket:

democracy/models/gdpr_data_serialization_mixin.py
nicobav: it is done.

democracy/models/gdpr_data_serialization_mixin.py
```diff
@@ -0,0 +1,14 @@
+from kerrokantasi.utils import get_current_request
+
+
+class FileFieldUrlSerializerMixin:
+    field_to_use_as_url_field = None  # Override in subclass.
+
+    @property
+    def url(self):
+        field = getattr(self, self.field_to_use_as_url_field, None)
```
charn, line 9: Since this is initialized with `None` already you can actually just do the following here:
```suggestion
        field =  self.field_to_use_as_url_field
```

charn (approved):
LGTM! :+1: :shipit:

PR #490, opened 2024-04-26
Fix crash when creating a hearing with existing file references, add option for creating new files/images by reference

danipran:
TL;DR:
- replace copy-or-update logic in CreateUpdateSerializers create-or-update
- fix crash when creating a hearing with existing file references (i.e. upload files before creating a hearing and then posting that hearing with the pre-uploaded file)
- add a write-only `reference_id` field for files/images when creating new hearings

---

Some of the serializers worked with a update-or-copy logic instead of update-or-create (as their name would imply), which was added in PR https://github.com/City-of-Helsinki/kerrokantasi/pull/363 (https://github.com/City-of-Helsinki/kerrokantasi/commit/6c97234277c1f07949858304f0060e0271e7e6e0).

This caused an error whenever you're creating a new hearing and have already uploaded a file to a section (via the form's file upload functionality). Added a test to confirm that this scenario functions correctly now. The crash is fixed (or should be fixed) in the PR's commit 32774cf664f9303570add11e7c9c1fc33604c5e1, but leaves the rather strange serializer logic in place. 

E.g. if you want to create a new hearing using an existing hearing as a base, you'll have to remove the IDs from the hearing and the sections, but *need* to have the IDs for files and images.

Since this works on an update-or-copy basis, it means that you *cannot* use an existing instance for a file or an image on creation; if you do, it gets copied. The only other way to use existing instances is to first create the hearing *without* the files/images and then update the hearing.

This PR replaces that copy (or rather, save as copy) logic with create logic (so create-or-update rather than copy-or-update) and moves the burden to the consumer of the API, i.e. the consumer needs to remove the IDs from the data they want to copy.

However, files and images still need a reference to the instance they're copied from, so add a new field "reference_id" for this. Not the perfect solution, but this allows the user to create new files or images based on existing instances while retaining some kind of sanity with the API.

In its current state, it's meant for "save as new" type of situations. This means that if the pk is set for an image/file, then it will ignore the reference_id, i.e. you cannot update an existing instance using reference_id.

Also fixes a crash when trying to get a URL for a file the user is trying to copy by adding a check for whether the file has a PK or not. E.g. try to copy a file -> get reference file from database -> set pk to `None` -> validation checks for URL -> URL requires a PK -> 💥

Also note that while the previous code referred to copying in comments and in its associated commit/PR, it's actually saving as a copy/saving as new rather than just copying. This is a **very** important distinction, since it means that you have the option to edit the hearing before creating a copy of it instead of just creating a clone of another hearing.

Refs KER-351

Related front-end PR: https://github.com/City-of-Helsinki/kerrokantasi-ui/pull/1025

### DIFF
--- a/democracy/views/section.py
+++ b/democracy/views/section.py
@@ -158,11 +158,7 @@ class SectionPollSerializer(serializers.ModelSerializer, TranslatableSerializer)
             serializer_params = {"data": option_data}
             if pk:
                 try:
-                    if self.instance.pk:
-                        option = self.instance.options.get(pk=pk)
-                    else:  # Creating a copy of the hearing
-                        option = SectionPollOption.objects.get(pk=pk)
-                        option.pk = None
+                    option = self.instance.options.get(pk=pk)
                 except SectionPollOption.DoesNotExist:
                     raise ValidationError("The Poll does not have an option with ID %s" % repr(pk))
                 serializer_params["instance"] = option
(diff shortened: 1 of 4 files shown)


terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr490.api.dev.hel.ninja** :rocket::rocket::rocket:

danipran:
Didn't go exactly as I thought. Back to the drawing board.

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr490.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr490.api.dev.hel.ninja** :rocket::rocket::rocket:

democracy/views/section.py
```diff
@@ -295,52 +291,64 @@ def update(self, instance, validated_data):
...
                 try:
-                    if self.instance is not None:
-                        # only allow orphan files or files within this section already
-                        file = SectionFile.objects.filter(Q(section=None) | (Q(section=self.instance))).get(pk=pk)
-                    else:  # Creating a copy of the hearing
-                        file = SectionFile.objects.get(pk=pk)
-                        file.pk = None
-                except SectionImage.DoesNotExist:
+                    # only allow orphan files or files within this section already
+                    file = SectionFile.objects.filter(Q(section=None) | (Q(section=self.instance))).get(pk=pk)
```
charn, line 335: A minor thing, but I think the extra parentheses that are wrapping `(Q(section=self.instance))` can be removed.

charn (approved):
Couldn't spot anything obvious, LGTM! :shipit:

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr490.api.dev.hel.ninja** :rocket::rocket::rocket:

PR #492, opened 2024-05-03
fix: copy files in copy_hearing function

danipran:
This PR makes the copy hearing action in Django Admin's hearing listing copy files as well.

Note that the hearings in Django Admin don't seem to work very well and I'm not even sure if it's used at all, but I encountered this bug while doing another change and fixed it. However, I decided to scrap the idea that used this function and made this fix into its own PR (rather than throwing it away).

Refs KER-359

### DIFF
--- a/democracy/models/utils.py
+++ b/democracy/models/utils.py
@@ -60,6 +60,12 @@ def copy_hearing(old_hearing, **kwargs):
             image.section = section
             image.save()
             _copy_translations(image, old_image)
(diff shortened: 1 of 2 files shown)


terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr492.api.dev.hel.ninja** :rocket::rocket::rocket:

charn (approved):
LGTM! :+1: :shipit:

PR #529, opened 2024-11-21
KER-399 | Command for removing old users

nicobav:
Adds functionality to remove users from objects (*_created, *_modified, *_deleted) and delete inactive users in form of django command.
Command to be added in cron job.

Refs KER-398 KER-399 KER-401

### DIFF
--- a/democracy/factories/comment.py
+++ b/democracy/factories/comment.py
@@ -1,5 +1,4 @@
 import random
-from datetime import timedelta
 
 import factory
 import factory.fuzzy
@@ -12,14 +11,10 @@ class BaseCommentFactory(factory.django.DjangoModelFactory):
     content = factory.Faker("text")
 
     @factory.post_generation
-    def post(self, create, extracted, **kwargs):
+    def create_random_voters(self, create, create_random_voters, **kwargs):
+        if not create_random_voters:
+            return
+
         for _x in range(int(random.random() * random.random() * 5)):
             self.voters.add(get_random_user())
         self.recache_n_votes()
-
-        # Can't be done in a lazy attribute because they have no access to the concrete factory's model class,  # noqa: E501
-        # for `parent_field`...
-        self.created_at = getattr(self, self.parent_field).created_at + timedelta(
-            seconds=random.randint(120, 600)
-        )
-        self.save(update_fields=("created_at",))
--- a/democracy/factories/hearing.py
+++ b/democracy/factories/hearing.py
@@ -84,9 +84,13 @@ class SectionFactory(factory.django.DjangoModelFactory):
         choices=SectionType.objects.exclude(identifier="main")
     )
     commenting = factory.fuzzy.FuzzyChoice(choices=Commenting)
+    hearing = factory.SubFactory(HearingFactory)
 
     @factory.post_generation
-    def post(self, create, extracted, **kwargs):
+    def create_random_comments(self, create, create_random_comments, **kwargs):
+        """By default, create random comments for the section."""
+        if not create_random_comments:
+            return
         for _x in range(random.randint(1, 5)):
             comment = SectionCommentFactory(section=self)
             logger.info(
--- a/democracy/tests/unittest/test_user_data_remover.py
+++ b/democracy/tests/unittest/test_user_data_remover.py
@@ -1,0 +1,369 @@
+import freezegun
+import pytest
+import reversion
+from django.utils import timezone
+from reversion.models import Version
+
+from democracy.factories.hearing import (
+    CommentImageFactory,
+    HearingFactory,
+    LabelFactory,
+    MinimalHearingFactory,
+    SectionCommentFactory,
+    SectionFactory,
+    SectionFileFactory,
+    SectionImageFactory,
+)
+from democracy.factories.organization import OrganizationFactory
+from democracy.factories.poll import SectionPollFactory, SectionPollOptionFactory
+from democracy.models import (
+    ContactPerson,
+    Project,
+    ProjectPhase,
+    SectionPollAnswer,
+    SectionType,
+)
+from democracy.utils import user_data_remover
+from kerrokantasi.models import User
+from kerrokantasi.tests.factories import UserFactory
+
+
+@pytest.fixture
+def user():
+    return UserFactory()
+
+
+@pytest.fixture
+def threshold_time():
+    return timezone.now() - timezone.timedelta(days=30)
+
+
+@pytest.mark.django_db
+class TestRemoveUserFromComments:
+    def test_user_data_nulled_after_threshold_time(self, user, threshold_time):
+        with freezegun.freeze_time(threshold_time - timezone.timedelta(days=1)):
+            comment = SectionCommentFactory(
+                created_by=user, section=SectionFactory(), flagged_by=user
+            )
+        dont_remove_user_from_comment = SectionCommentFactory(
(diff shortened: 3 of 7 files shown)


terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr529.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr529.api.dev.hel.ninja** :rocket::rocket::rocket:

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr529.api.dev.hel.ninja** :rocket::rocket::rocket:

democracy/factories/comment.py
```diff
@@ -13,6 +13,8 @@ class BaseCommentFactory(factory.django.DjangoModelFactory):
 
     @factory.post_generation
     def post(self, create, extracted, **kwargs):
+        if extracted:
+            return
```
danipran, line 17: I think I get the idea here (you want to disable the post-hook from running if `post` has any value set), but it's very confusing... and so is the original implementation. How this should look like is something like this:

```py
# First of all, just remove the BaseCommentFactory and fully implement SectionCommentFactory instead
class SectionCommentFactory(factory.django.DjangoModelFactory):
    created_by = factory.LazyAttribute(lambda obj: get_random_user())
    content = factory.Faker("text")

    class Meta:
        model = SectionComment

    @factory.post_generation
    def set_created_at_in_future(self, create, extracted, **kwargs):
        # NOTE: I really hate keeping this part of the code, but I've no clue if removing this would break any tests or not. Remove if not.
        # Can't be done in a lazy attribute because they have no access to the concrete factory's model class,  # noqa: E501
        # for `parent_field`...
        self.created_at = getattr(self, self.parent_field).created_at + timedelta(
            seconds=random.randint(120, 600)
        )
        self.save(update_fields=("created_at",))


class SectionCommentWithVotes(SectionCommentFactory):
    @factory.post_generation
    def add_random_voters(self, create, extracted, **kwargs):
        # NOTE: This is terrible code and makes the tests *not* deterministic, but I'm 99% sure removing this would break a lot of tests.
        for _x in range(int(random.random() * random.random() * 5)):
            self.voters.add(get_random_user())
        self.recache_n_votes()

# In a normal, sane test...
def test_foo():
    my_comment = SectionCommentFactory(...)
    ...

# In every existing test in this project:
def test_bar():
    my_comment_with_votes = SectionCommentWithVotesFactory(...)
    ...
```

Of course, this solution requires you to rename a lot of things (namely, `SectionCommentFactory` -> `SectionCommentWithVotesFactory`, don't need to deal with variable names etc.). However, it also makes it more clear what's going on in the future. I think it's worth doing.

democracy/factories/hearing.py
```diff
@@ -84,9 +84,12 @@ class Meta:
         choices=SectionType.objects.exclude(identifier="main")
     )
     commenting = factory.fuzzy.FuzzyChoice(choices=Commenting)
+    hearing = factory.SubFactory(HearingFactory)
 
     @factory.post_generation
     def post(self, create, extracted, **kwargs):
+        if extracted:
+            return
```
danipran, line 92: Same thing here as with `BaseCommentFactory`, except you don't have to deal with a base class. Separate this into two different factories, `SectionFactory` (same as the current one, but without the post-generation hook) and `SectionWithCommentsFactory` (subclass of `SectionFactory` with the post-generation hook from the original version)

danipran (changes requested):
I think the functionality itself looks fine, but I have some suggestions for the tests and some nitpicking on names/typing.

democracy/utils/user_data_remover.py
```diff
@@ -0,0 +1,200 @@
...
+
+
+def get_inactive_users(
+    threshold_datetime: datetime,
+) -> QuerySet:
+    """Returns all users that have not created or modified hearings, comments, polls,
+    organizations, files, images, polls or voted since the given threshold time.
+    """
+    return User.objects.filter(date_joined__lte=threshold_datetime).exclude(
+        Q(sectioncomment_created__isnull=False)
```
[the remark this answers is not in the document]
nicobav: I agree. The exclusion was left from some old iteration of the code. And actually I'm sure we can ditch those `Q`'s as well with the `filter(`

democracy/factories/comment.py
nicobav: Yes, the refactoring would pay itself in the future. In the process of fighting with the tests the idea of refactoring bunch of unrelated tests seemed a deeper pithole.

democracy/factories/hearing.py
nicobav: I think adding new simpler factories would perhaps be better way to operate this.

democracy/tests/unittest/test_user_data_remover.py
```diff
@@ -0,0 +1,358 @@
...
+@pytest.mark.django_db
+class RemoveUserTestsBase:
+    @pytest.fixture(autouse=True)
+    def init_test_data(self):
+        self.threshold_time = timezone.now() - timezone.timedelta(days=30)
+        self.user = UserFactory(username="user")
+
+
+class TestRemoveUserFromComments(RemoveUserTestsBase):
+    def test_user_data_nulled_after_threshold_time(self):
```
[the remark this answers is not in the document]
nicobav: oh, I see this. I think the author had a the days settings in mind when naming this.

democracy/utils/user_data_remover.py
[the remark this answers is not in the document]
nicobav: The docs needs to be changed also since it has description of the old iteration of this function.

democracy/tests/unittest/test_user_data_remover.py
```diff
@@ -0,0 +1,358 @@
...
+from kerrokantasi.models import User
+from kerrokantasi.tests.factories import UserFactory
+
+
+@pytest.mark.django_db
+class RemoveUserTestsBase:
+    @pytest.fixture(autouse=True)
+    def init_test_data(self):
+        self.threshold_time = timezone.now() - timezone.timedelta(days=30)
+        self.user = UserFactory(username="user")
```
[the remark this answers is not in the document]
nicobav: Imo the variable / fixture naming matters - where it is initialized don't. Since if you want to know what the fixture has been eaten you still need to travel your IDE to the point of creation. But what comes to base class that's a bit overkill here (pros for defining variables and annotations once but that a loose pro and cons it clearly can obfuscate things).
As conclusion I'm agreeing you with my comments in mind and I'll remove the base class and change these vars to fixtures because this case it seems clearer 👍

democracy/factories/hearing.py
nicobav: for now I thought just get ridoff the `BaseCommentFactory`'s random creation time functionality and add proper names for the post hooks and give possibility to exclude them.

danipran: Yeah, I get it... this is just something that won't ever get done. I'll make a ticket out of this, maybe the legendary someone™ will eventually finish the job. Most likely not.

democracy/factories/comment.py
danipran: Will make a ticket out of this.

danipran (approved):

PR #542, opened 2025-03-24
chore: whitelist pyproject.toml

danipran:
`pyproject.toml` contains configurations for pytest, for example. This needs to be included in the container or else some features might not work as intended.


nicobav (approved):
Nicccce

PR #559, opened 2025-06-13
Change logging format to json.

JokinenK:
Install and configure django-logger-extra library and use it to implement structured logging in json format.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -203,6 +203,10 @@ charset-normalizer==3.3.2 \
     --hash=sha256:fd1abc0d89e30cc4e02e4064dc67fcc51bd941eb395c502aac3ec19fab46b519 \
     --hash=sha256:ff8fa367d09b717b2a17a052544193ad76cd49979c805768879cb63d9ca50561
     # via requests
+colorama==0.4.6 \
+    --hash=sha256:08695f5cb7ed6e0531a20572697297273c47b8cae5a63ffc6d6ed5c201be6e44 \
+    --hash=sha256:4f1d9991f5acc0ca119f9d443620b77f9d6b33703e51011c16baf57afb285fc6
+    # via ipython
 cryptography==44.0.1 \
     --hash=sha256:00918d859aa4e57db8299607086f793fa7813ae2ff5a4637e318a25ef82730f7 \
     --hash=sha256:1e8d181e90a777b63f3f0caa836844a1182f1f265687fac2115fcf245f5fbec3 \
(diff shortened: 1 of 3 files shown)


voneiden (changes requested):
Commit message needs to be in conventional commit format, so you can change that to for example `feat: change logging format to json`

The deps are a bit flaky depending on OS/architecture, looks like in your case it has decided to drop `pexpect` and `ptyprocess` and replace them with `colorama`. Was this a Windows/WSL thing? Anyway, gotta watch out for that.

JokinenK:
> Commit message needs to be in conventional commit format, so you can change that to for example `feat: change logging format to json`
> 
> The deps are a bit flaky depending on OS/architecture, looks like in your case it has decided to drop `pexpect` and `ptyprocess` and replace them with `colorama`. Was this a Windows/WSL thing? Anyway, gotta watch out for that.

Yes, it is Windows/WSL setup. I can boot into linux to correct this, did not expect WSL to act differently compared to native.

requirements.txt
```diff
@@ -355,7 +359,6 @@ djangorestframework==3.15.2 \
     --hash=sha256:36fe88cd2d6c6bec23dca9804bab2ba5517a8bb9d8f47ebc68981b56840107ad
     # via
     #   -r requirements.in
-    #   django-helsinki-notification
```
voneiden, line 358: This removed line should be still present...? 

```bash
$ pip-compile --generate-hashes --strip-extras requirements.in
$ git diff
``` 


```diff
diff --git a/requirements.txt b/requirements.txt
index d34b99d..b3b9baa 100644
--- a/requirements.txt
+++ b/requirements.txt
@@ -359,6 +359,7 @@ djangorestframework==3.15.2 \
     --hash=sha256:36fe88cd2d6c6bec23dca9804bab2ba5517a8bb9d8f47ebc68981b56840107ad
     # via
     #   -r requirements.in
+    #   django-helsinki-notification
     #   django-parler-rest
     #   drf-nested-routers
     #   drf-oidc-auth
```

voneiden (changes requested):

terovirtanen:
**KERROKANTASI-API branch is deployed to platta: https://kerrokantasi-pr559.api.dev.hel.ninja** :rocket::rocket::rocket:

requirements.txt
JokinenK: Running the same command, I do receive warning and it does not add that given line:
`WARNING: django-helsinki-notification 0.0.0 does not provide the extra 'rest_framework'`

JokinenK: Ended up being problem with pip 24.0. Updated to 25.1 fixed the issue and restored that missing line. Too new pip on the other hand broke pip-tools compatibility.

voneiden (approved):
To summarize here as well, had some compatibility issues which were resolved with pip 25.0, however 25.1 is again incompatible with pip-tools.

PR #661, opened 2026-08-12
chore(deps): make django dep lower case

nicobav:
To fix issues updating django dependency with dependabot make the django lowercase.

Refs: KER-633

## Summary by CodeRabbit

* **Chores**
  * Standardized the Django dependency name while retaining the existing version requirement.


mhieta (approved):
LGTM!

